PR #4, opened 2022-02-26
Implement PyPI publishing

ConorMacBride:
The built matrix [can be parsed from JSON](https://docs.github.com/en/actions/learn-github-actions/expressions#example-returning-a-json-object) so having a separate job to create the build matrix should work. I experimented with allowing for more configuration of the CIBW env vars by parsing JSON, however, in my opinion it adds a lot of unnecessary complexity. In most cases cibuildwheel can be [configured in `pyproject.toml`](https://cibuildwheel.readthedocs.io/en/stable/options/#configuration-file). Using the configuration file would mean less to maintain in this repo.

At some point the `generate_platform_matrix` job may need modified if we want to allow for `aarch64` builds with the `hypriot/qemu-register` Docker image.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,16 +18,33 @@ on:
         required: false
         default: false
         type: boolean
+      platforms:
+        description: List of target platforms
+        required: false
+        default: 'ubuntu-20.04 windows-2019 macos-10.15'
+        type: string
 
 jobs:
 
+  generate_platform_matrix:
+    name: Generate build_wheels matrix
+    runs-on: ubuntu-latest
+    if: ${{ !inputs.pure_python_wheel }}
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - id: set-matrix
+        run: |
+          FILTER='split(" ") as $platforms | {"os": $platforms}'
+          echo "::set-output name=matrix::$( echo ${{ inputs.platforms }} | jq -R -c $FILTER )"
+
   build_wheels:
     name: Build platform-dependent wheels
+    needs: [generate_platform_matrix]
     runs-on: ${{ matrix.os }}
     if: ${{ !inputs.pure_python_wheel }}
     strategy:
-      matrix:
-        os: [ubuntu-20.04, windows-2019, macos-10.15]
+      matrix: ${{fromJSON(needs.generate_platform_matrix.outputs.matrix)}}
     steps:
     - uses: actions/checkout@v2
       with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,6 +23,19 @@ on:
         required: false
         default: 'ubuntu-20.04 windows-2019 macos-10.15'
         type: string
+      upload_to_pypi:
+        description: Always upload to PyPI - if not, only upload tags
+        required: false
+        default: false
+        type: boolean
+      repository_url:
+        description: The PyPI repository URL to use
+        required: false
+        default: ''
+        type: string
+    secrets:
+      pypi_password:
+        required: false
 
 jobs:
 
@@ -56,6 +69,9 @@ jobs:
         env:
           CIBW_TEST_EXTRAS: ${{ inputs.test_extras }}
           CIBW_TEST_COMMAND: ${{ inputs.test_command }}
+      - uses: actions/upload-artifact@v2
+        with:
+          path: dist/*
 
   build_sdist:
     name: Build source distribution
@@ -68,6 +84,9 @@ jobs:
         with:
           test_extras: ${{ inputs.test_extras }}
           test_command: ${{ inputs.test_command }}
+      - uses: actions/upload-artifact@v2
+        with:
+          path: dist/*
 
   build_sdist_and_wheel:
     name: Build source and wheel distribution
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -14,52 +14,128 @@ on:
...
+    name: Generate build_wheels matrix
+    runs-on: ubuntu-latest
+    if: ${{ !inputs.pure_python_wheel }}
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - id: set-matrix
+        run: |
+          FILTER='split(" ") as $platforms | {"os": $platforms}'
+          echo "::set-output name=matrix::$( echo ${{ inputs.platforms }} | jq -R -c "$FILTER" )"
```
Cadair, line 57: Oh well!

ConorMacBride: I've now found a much better way of doing this by using a Python script to parse a YAML string and return a build matrix as a JSON string! Was experimenting [here](https://github.com/ConorMacBride/test-workflows) and will update this PR now. Basically, it recreates the ["targets:" list](https://openastronomy-azure-pipelines.readthedocs.io/en/latest/publish.html#usage-and-options) from the old template exactly.

.github/workflows/publish.yml
```diff
@@ -13,53 +17,136 @@ on:
...
+        required: false
 
 jobs:
 
+  targets:
+    name: Load build targets
+    runs-on: ubuntu-latest
+    outputs:
+      sdist: ${{ steps.set-outputs.outputs.sdist }}
+      universal: ${{ steps.set-outputs.outputs.universal }}
```
astrofrog, line 46: We should avoid the term universal which was a mistake in the original templates (universal means pure Python but also compatible with Python 2 and 3). Maybe we can call this pure_python instead?

ConorMacBride: Okay, I'll change that back.

ConorMacBride:
This now takes the same list of `targets` as the Azure Pipelines publish template. (It's working except I need to find out how to build for aarch64.)

I'll need to work out how to [install libraries for testing sdist](https://github.com/OpenAstronomy/azure-pipelines-templates/pull/41/files). For cibuildwheel they are installed with the [`before-all` option](https://cibuildwheel.readthedocs.io/en/stable/options/#before-all). I should be able to add these libraries to the build matrix with the same python script, and then iterate over them in a bash script.

.github/workflows/publish.yml
```diff
@@ -13,53 +17,136 @@ on:
...
+      - uses: actions/checkout@v2
+# Uncomment before merging
+#        with:
+#           repository: 'OpenAstronomy/github-actions-workflows'
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+      - run: python -m pip install PyYAML
+      - id: set-outputs
+        run: python tools/load_build_targets.py "${{ inputs.targets }}"
```
Cadair, line 58: This really is an amazing workaround! :clap: 

Dear Github actions: you are missing a critical feature.

astrofrog:
This is looking amazing!

I wonder whether we should split the workflow for pure Python packages into a separate workflow as it would then mean minimal configuration for them. 

We could then make the workflow here default to sdist plus all common platform wheels.

This would mean that packages should be able to not specify targets and this would be reserved for more advanced use cases?

Cadair:
> We could then make the workflow here default to sdist plus all common platform wheels if target is not specified.

Why don't we default to be a pure python only in this workflow and then if people need binary wheels then they have to specify targets? We just need to figure out a way to fix https://github.com/OpenAstronomy/build-python-dist/issues/2 and then it would throw an error if a project which needs to build platform specific stuff didn't configure it?

Additionally we could make "wheel" default to win/mac/linux py[37-310] amd64? in the target list? Although I am kinda of the opinion if you are building binary wheels opting in explicitly to the platforms isn't a bad thing?

ConorMacBride:
> I wonder whether we should split the workflow for pure Python packages into a separate workflow as it would then mean minimal configuration for them.

We could move the configuration for these back out of the Python script so the Python script only selects the cibuildwheel targets?

How about the following behaviour:

```
# sdist and pure python wheel
uses: publish.yml
  with:
    pure_python: wheel  # default
```

```
# sdist only
uses: publish.yml
  with:
    pure_python: sdist
```

```
uses: publish.yml
  with:
    pure_python: sdist  # default if targets specified
    targets: |
      - wheels_linux  # run on ubuntu and let cibuildwheel decide what to build
      - wheels_macos  # run on macOS ...
      - wheels_windows  # run on windows ...
      - wheels_<CIBW_BUILD>
```

Or maybe it would make more sense to separate the sdist and pure python wheel options with `sdist: [true|false]` and `pure_python: [true|false]`?

Also, should the `build_sdist` and `build_sdist_and_wheel` jobs be merged as they contain the same steps?

astrofrog:
I think pure_python: sdist is confusing as the package with extensions would not be pure Python. I still think having two workflows would be nicer and the pure Python one would be super simple and easy to maintain. After all they are conceptually two different workflows.

astrofrog:
If you prefer to keep as one workflow then I think a single pure_python: true/false defaulting to true makes sense, with sdist listed in the targets if desired.

ConorMacBride:
Yeah, two workflows would work well. Each workflow would contain the publish to PyPI step which should be fine.

.github/workflows/publish.yml
```diff
@@ -13,53 +26,120 @@ on:
...
+    name: Load build targets
+    if: ${{ inputs.targets != '' }}
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-outputs.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+# Uncomment before merging
+#        with:
+#           repository: 'OpenAstronomy/github-actions-workflows'
```
ConorMacBride, line 65:
```suggestion
        with:
           repository: 'OpenAstronomy/github-actions-workflows'
```

ConorMacBride (commented):
Except the the final suggested change, **I think this is ready to be merged, provided you're happy with the API?**

I will open separate issues to discuss other improvements, such as sphinx documentation and more robust testing, as well as the missing features from the Azure Pipelines template:

- Ability to publish npm packages (for Jupyter widget packages)
- Needs option to remove local scheme from version (with [execute-517](https://github.com/Cadair/execute-517)) (for dev releases)
- Test and document publishing to e.g. Azure Artifacts (how to configure authentication)

I went through the Azure template line-by-line and all the documented (and undocumented) behaviours and features have either been implemented or noted above.

.github/workflows/publish_pure_python.yml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          lfs: true
+          submodules: ${{ inputs.submodules }}
+      - name: Install dependencies
+        if: ${{ inputs.libraries != '' }}
+        uses: ConorMacBride/install-package@main
```
Cadair, line 53: I am happy to move this under OpenAstronomy as well if you want some help looking after it.

Cadair (approved):

Cadair:
I am happy with the interface, merge it and let's test it on some other repos!

PR #5, opened 2022-03-01
Implement tox testing

ConorMacBride:
Similar to the publish workflow, a Python script is used to create the matrix of tox environments. While it's not ideal to have to use a Python script to configure the CI pipeline, it does offer a lot of flexibility and makes it much easier to implement new options. We could even test this Python script itself.

I've create a composite action (https://github.com/ConorMacBride/install-package) to replace https://github.com/OpenAstronomy/azure-pipelines-templates/blob/master/install-libraries.yml. It's used in both the publish and tox workflows. This can be moved to the OpenAstronomy org if you like.

Have a look at `test_tox.yml` and let me know what you think of how the workflow is configured. I've tried to keep it similar to the Azure Pipelines tox template, but there are still a lot of key features to add -- including actually running tox!

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -17,6 +17,11 @@ on:
         required: false
         default: ''
         type: string
+      toxdeps:
+        description: Tox dependencies
+        required: false
+        default: ''
+        type: string
       toxargs:
         description: Positional arguments for tox
         required: false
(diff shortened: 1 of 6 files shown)


.github/workflows/tox.yml
```diff
@@ -0,0 +1,142 @@
...
+  envs:
+    name: Load tox environments
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-outputs.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      # Uncomment before merging
+      #        with:
+      #           repository: 'OpenAstronomy/github-actions-workflows'
```
ConorMacBride, line 76:
```suggestion
        with:
          repository: 'OpenAstronomy/github-actions-workflows'
```

ConorMacBride (commented):
This is ready to merge, unless you have any suggestions?

astrofrog:
I tried using this in the wild on extension-helpers  (see https://github.com/astropy/extension-helpers/pull/37) and it looks like this doesn't load the ``tox_matrix.py`` script from the right location:

```
python: can't open file '/home/runner/work/extension-helpers/extension-helpers/tools/tox_matrix.py': [Errno 2] No such file or directory
```

ConorMacBride:
That's because for testing in this PR it is checking out the current ref rather than the main branch of this repo. See the "Uncomment before merging" code suggestion above.

ConorMacBride:
Should we just merge this PR and make changes later if needed?

astrofrog:
Yes, feel free to go ahead and merge!

PR #13, opened 2022-03-10
Update how upload_to_pypi is defined

astrofrog:
Prior to this PR, I think upload_to_pypi was potentially counter-intuitive because upload_to_pypi: false still means that sometimes things are uploaded to PyPI. I think it would be cleaner to simply specify that it is a boolean condition for the upload, defaulting to true if a v* tag is pushed.

I'm not sure if this will work though, as I don't know if a condition can be given as the default for the parameter. But I do think we should try and make it behave somehow like it is here, meaning that upload_to_pypi: false would never upload anything.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -144,7 +144,7 @@ jobs:
     if: |
       always() &&
       needs.targets.result == 'success' &&
-      needs.targets.outputs.upload_to_pypi &&
+      needs.targets.outputs.upload_to_pypi == 'true' &&
       needs.build_wheels.result != 'failure' &&
       needs.build_sdist.result != 'failure'
     steps:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -32,9 +32,9 @@ on:
         default: ''
         type: string
       upload_to_pypi:
-        description: Always upload to PyPI after successful builds - if false, only upload when pushing tags
+        description: A condition specifying whether to upload to PyPI
         required: false
-        default: false
+        default: (github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v'))
         type: boolean
       repository_url:
         description: The PyPI repository URL to use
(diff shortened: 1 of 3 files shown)


astrofrog:
Should we make the parameter a string and then somehow evaluate it as a boolean later in the pipeline for it to work?

astrofrog:
@ConorMacBride - I probably won't have time to try and dig deeper into this today so feel free to take over the PR if you like, if you agree that we should tweak this option.

ConorMacBride:
A string would work. I can take a look later. I'll make `upload_to_pypi` accept `'true'`, `'false'` and `'auto'`. Or should we call it `'tag'` instead of `'auto'`?

astrofrog:
I guess the question is how would a user then specify that they want to upload on all tags, not just ones starting with 'v'? Basically we need a way for users to pass a completely custom condition?

ConorMacBride:
We could keep `upload_to_pypi` a boolean input and default to `false`. When the reusable workflow is called, it should be possible to do
```yaml
with:
  upload_to_pypi: ${ github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v') }
```

ConorMacBride:
If we do that I think we should make `upload_to_pypi` be required (no default) to reduce confusion (e.g. why does the publish workflow not publish by default?)

astrofrog:
@ConorMacBride yes we might have to just make it required. @Cadair, what do you think?

ConorMacBride:
I've updated the `upload_to_pypi` input to have the following behaviour. (With some testing here: https://github.com/ConorMacBride/test-workflows/blob/main/.github/workflows/test_upload.yml)

#### upload_to_pypi
Whether to upload to PyPI after successful builds.
The default is to upload to PyPI when tags that start with `v` are pushed.
A boolean can be passed as `true` (always upload) or `false` (never upload)
either explicitly or as a boolean expression (`${{ <expression> }}`).

Alternatively, a string can be passed to match the start of a tag ref.
For example, `'refs/tags/v'` (default) will upload tags that begin with `v`,
and `'refs/tags/'` will upload on all pushed tags.

astrofrog:
Nice! Thank you!

ConorMacBride (approved):
Merge?

astrofrog:
@ConorMacBride - could you rebase to resolve the conflict? Then feel free to merge.

PR #15, opened 2022-03-10
Add env vars for changing runner images

ConorMacBride:
This adds optional environment variables `OA_LINUX_RUNNER`, `OA_MACOS_RUNNER` and `OA_WINDOWS_RUNNER` which, if set, override the default `-latest` images. The environment variables need to be passed to the reusable workflow directly. E.g.:
```yaml
jobs:
  test:
    with:
      envs: |
        - macos: py310
        - windows: py39-docs
    env:
      OA_WINDOWS_RUNNER: windows-2019
```

xref https://github.com/astropy/extension-helpers/pull/37

Do you think the env var solution is good enough?

### DIFF
--- a/tools/load_build_targets.py
+++ b/tools/load_build_targets.py
@@ -5,9 +5,9 @@ import click
 import yaml
 
 MACHINE_TYPE = {
-    "linux": "ubuntu-latest",
-    "macos": "macos-latest",
-    "windows": "windows-latest",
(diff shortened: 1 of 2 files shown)


astrofrog (approved):
Yes I think this is probably fine, thanks!

ConorMacBride:
I think we'll need to move the env variables to actual inputs as apparently the env context isn't supported for reusable workflows. I misread: https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations

PR #16, opened 2022-03-10
Add tox workflow input for choosing runner images

ConorMacBride:
This reverts the changes made in https://github.com/OpenAstronomy/github-actions-workflows/pull/15 and adds the following new option to the tox workflow only:

#### runner
Choose an alternative image for the runner to use on each OS.
By default, `linux` is `ubuntu-latest`, `macos` is `macos-latest` and `windows` is `windows-latest`.
None, some or all OSes can be specified, and the global value can be overridden in each environment.
It can be defined globally:
```yaml
with:
  runner: |
    linux: ubuntu-18.04
    macos: macos-10.15
    windows: windows-2019
```
```yaml
with:
  runner: |
    macos: macos-10.15
```
***Note:** `runner` is a **string** and must be specified as a literal block scalar using the `|`. (Without the `|`, it must also be valid YAML.)*
`envs` definition:
```yaml
with:
  envs: |
    - windows: py39
      runner: windows-2019
```

### DIFF
--- a/.github/workflows/test_tox.yml
+++ b/.github/workflows/test_tox.yml
@@ -11,6 +11,7 @@ jobs:
         - macos: py39-inputs-macos
         - windows: py38-inputs-windows
           toxargs: '-v'
+          runner: windows-2019
 
         - linux: py310-inputs-conda
         - macos: py310-inputs-conda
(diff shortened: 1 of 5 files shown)


ConorMacBride:
This time it works! Is `runner` the best name? Was thinking it could be added like
```yaml
envs: |
  - windows-2019: py39
```
But I think that would add unnecessary complexity. Using this instead:
```yaml
envs: |
  - windows: py39
    runner: windows-2019
```

astrofrog:
The last example you sent is good. What about just using runs-on as the name?

ConorMacBride:
Changed to `runs-on`, and the tests are selecting the image correctly. Merge?

astrofrog (approved):
Looks great!

PR #17, opened 2022-03-10
Simplify pure Python workflow into a single job

astrofrog:
As mentioned by @pllim in https://github.com/astropy/pytest-astropy-header/pull/42#issuecomment-1064204463, for the pure Python case we could just have a single job?

### DIFF
--- a/.github/workflows/publish_pure_python.yml
+++ b/.github/workflows/publish_pure_python.yml
@@ -59,21 +59,9 @@ jobs:
           test_extras: ${{ inputs.test_extras }}
           test_command: ${{ inputs.test_command }}
           pure_python_wheel: true
-      - uses: actions/upload-artifact@v2
-        with:
-          path: dist/*
-
-  upload_pypi:
-    name: Upload to PyPI
-    needs: [build]
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


pllim (approved):

.github/workflows/publish_pure_python.yml
```diff
@@ -59,21 +59,9 @@ jobs:
           test_extras: ${{ inputs.test_extras }}
           test_command: ${{ inputs.test_command }}
           pure_python_wheel: true
-      - uses: actions/upload-artifact@v2
-        with:
-          path: dist/*
```
ConorMacBride, line 64: Would it be useful to keep the upload-artefact so people can download the wheel even if it isn't uploaded to PyPI? (Also, it should be easy to rebase https://github.com/OpenAstronomy/github-actions-workflows/pull/13 for this PR.)

ConorMacBride: Although happy to merge as I guess this is less important for pure python wheels.

pllim: People can always build their own wheel but why would they even need it?

ConorMacBride: That's right. Building locally to debug the wheels should be fine. Shall we merge this?

ConorMacBride (approved):

PR #18, opened 2022-03-11
Rename pypi_password to pypi_token

astrofrog:
As suggested on Matrix - I think we should probably make this change sooner rather than later to avoid breaking things?

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -47,7 +47,7 @@ on:
         default: true
         type: boolean
     secrets:
-      pypi_password:
+      pypi_token:
         required: false
 
 jobs:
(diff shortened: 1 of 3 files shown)


ConorMacBride (approved):

astrofrog:
@Cadair - do you know if PyPI repositories other than the main one accept tokens?

Cadair:
Not a clue

astrofrog:
Could we just have pypi_token and then if we need in future we cold add pypi_username and pypi_password for cases where tokens aren't possible, if that is needed?

astrofrog:
Then we could just set the username to __token__ by default and then for the password we can use 'password' if the username is not __token__ and otherwise pypi_token?

ConorMacBride:
That would be possible, although unlike Azure, GitHub Actions doesn't seem to have a `coalesce(pypi_password, pypi_token)` function so might take a bit of work. Easiest option might be a separate upload action depending on if password or token is set.

astrofrog:
Ok sounds good - I do think for the default case, token makes more sense (as the fact that the token is set as the password is kind of an implementation detail)

astrofrog:
Ok sounds good - let's merge :)

Issue #9, opened 2022-03-02
Reference the reusable workflow repo inside the workflow


ConorMacBride:
To run the Python scripts within the reusable workflows we checkout the `main` branch of this repository. However, this means that if a PR updates a Python script the script that the PR CI tests is the version in the `main` branch and no the PR branch. Need to change the following in each workflow:

```yaml
- uses: actions/checkout@v2
  with:
    repository: 'OpenAstronomy/github-actions-workflows'
```

ConorMacBride:
Actually, the PR will test the Python script from the PR branch because `'OpenAstronomy/github-actions-workflows'` is the default value and in that case the current ref is used.

However, if the reusable workflow is called by a different repository it will checkout the default branch of `OpenAstronomy/github-actions-workflows` **and not e.g. `refs/tags/v1` if the workflow was referenced as `...tox.yml@v1`**. See https://github.community/t/reusable-workflows-get-the-ref-inside-the-called-workflow/224109

We could ask the user to pass the ref they used as an input to the reusable workflows although that is not ideal. Hopefully there will be a better solution soon.

astrofrog:
An ugly solution would be that when we are ready to tag, we add a commit that hard-codes the ref of the repo to check out, e.g. ``v1.0.2``, and then tag that commit ``v1.0.2``, then add a commit that removes the ref again. I think it would work, although is pretty ugly.

astrofrog:
Another solution, not very elegant but which would at least work would be to embed the Python scripts inside the workflow YAML, for instance:

```echo "
<contents of tox_matrix.py>
" > tox_matrix.py
python tox_matrix.py ...
```

Yes this is ugly but it might be that the Python script doesn't need to be updated too often so it might not be *that* bad.

astrofrog:
See https://github.com/OpenAstronomy/github-actions-workflows/pull/20 for a proof of concept

astrofrog:
@ConorMacBride should we close this?

ConorMacBride:
This is no longer an issue so maybe just close for less noise. But we should switch to a solution that doesn't involve duplication if/when GitHub adds such a feature.

ConorMacBride:
https://github.com/actions/toolkit/issues/1264 seems to suggest it's still not possible via the `github` context. I'm guessing the `github.workflow_ref` might refer to the workflow that calls the reusable workflow. Looking through some of the workarounds mentioned in the issue we should be able to use the GitHub API.

[...]

PR #20, opened 2022-03-13
Embed python script in YAML file

astrofrog:
Proof of concept of a possible solution to https://github.com/OpenAstronomy/github-actions-workflows/issues/9

In practice I think if we really wanted to do this we should add a pre-commit hook to make sure the base64 encoded script is in sync with the Python file in the repo, and also to run in CI. We could also add a flag to the script to make it return a base64-encoded version of itself.

This is hacky but I think would actually solve https://github.com/OpenAstronomy/github-actions-workflows/issues/9 🤷 

(this only updates one of the scripts)


astrofrog:
To be clear, I am suggesting this as a temporary option until we can properly access the workflow ref inside the workflow. It would at least allow us to start tagging some versions.

ConorMacBride:
This is the first time I've seem a whole program encoded into an environment variable! This does seem like the best solution for now. The GitHub Actions docs advises on [Storing Base64 binary blobs as secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets#storing-base64-binary-blobs-as-secrets). Should be do: `echo $TOX_MATRIX_SCRIPT | base64 --decode > tox_matrix.py` instead of using python to decode? Use https://pre-commit.com for the pre commit hook?

astrofrog:
Note that we can't use secrets as they won't be frozen for each tag. I can try and go ahead and implement this properly.

Cadair:
How about we only do this in our test workflow, and if the env var isn't defined in the main workflow then it reads from the file?

ConorMacBride:
> How about we only do this in our test workflow, and if the env var isn't defined in the main workflow then it reads from the file?

I'm not sure what you mean by this. If it reads the python script from file we have no way of knowing which version/ref of the file it should read.

astrofrog:
@Cadair - but we would be defining TOX_MATRIX_SCRIPT in the workflow file like in this PR so it would always be defined?

astrofrog:
Ok so now I need to learn about how pre-commit works unless someone fancies just pushing a commit here 😆

ConorMacBride:
I can take a look later. Do you want me to configure https://pre-commit.ci so we can auto-update the encoded script in PRs?

Cadair:
I have added the pre-commit CI

ConorMacBride (approved):

Cadair (approved):

PR #22, opened 2022-03-15
Move workflows to root of repo

### LINKED ISSUE #21: Clean up workflow file path
Is there a way we can change this:

```
```

into this:
```
uses: OpenAstronomy/github-actions-workflows/tox.yml@main
```

Having the `.github` in particular is ugly, but also just making this shorter is going to make things better for a lot of downstreams.

ConorMacBride:
For testing if moving the workflows to the root actually works. If this works, should also test that it works from an external repository.

Closes #21


ConorMacBride:
Yes

PR #30, opened 2022-03-18
Fix pytest flags

### LINKED ISSUE #27: Adjust README to mention v1 tag?
Should the examples in the README use the @v1 tag?

### LINKED ISSUE #29: Error for packages that don't have pytest-cov as a dependency
See https://github.com/astropy/pytest-astropy-header/runs/5602096387?check_suite_focus=true - for an example of a failure because the package doesn't install pytest-cov.

ConorMacBride:
Only add the `--cov-report` flag when the `coverage` input contains `codecov`. Fixes #29 

Replaces `main` with `v1` in README. Closes #27 

Also, 

- removed `--junitxml` from pytest flags. (This option was needed in Azure Pipelines for displaying the test results in the Azure UI but GitHub Actions does not do this.)
- added .gitignore taken from OpenAstronomy/azure-pipelines-templates.


ConorMacBride:
pre-commit.ci autofix

ConorMacBride:
The autofix doesn't work as I would expect. If it says "files were modified by this hook" I would expect it to simply commit those modifications.

astrofrog (approved):

PR #32, opened 2022-03-31
DOC: Add upload_to_pypi example

### LINKED ISSUE #31: DOC: Working example of upload_to_pypi tag customization
It is explained in README that it can be customized but would be nice to have an example on how to drop the leading "v". Some packages tag `X.Y.Z` instead of `vX.Y.Z`. Thanks!

https://github.com/OpenAstronomy/github-actions-workflows/blob/ce411ebef92d6048328bee7486f0e2c8de1bece4/.github/workflows/publish_pure_python.yml#L21-L24

cc @tepickering

pllim:
Fix #31


ConorMacBride (approved):
Thanks!

PR #35, opened 2022-04-01
Add `fail-fast` input to `tox.yml` and `publish.yml`

### LINKED ISSUE #33: Configuring fast failing
At the moment it seems that by default the fast-fail strategy is used which means that all builds are cancelled if one fails. It would be nice to have an option to change this to not fast-fail.

ConorMacBride:
This adds a `fail-fast` input to the tox and cibuildwheel publish templates. This is passed directly to [`fail-fast`](https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs#canceling-remaining-jobs-if-a-matrix-job-fails) in the job stratagy.

#### Default values

For the tox template, I have set it to not fail fast by default because people typically want to see the results for all the tests within a single call to the tox workflow. (And  this is what the Azure template does.) However, this goes against GitHub's default which is `fail-fast: true`.

For the publish template, I have kept the GitHub default option. As this template is more intended for publishing (rather than testing) it would make more sense to give up as soon as possible, and within each job cibuildwheel fails fast anyway.

Does that seem sensible?

Closes #33

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -41,6 +41,11 @@ on:
         required: false
         default: ''
         type: string
+      fail-fast:
+        description: Whether to cancel all in-progress jobs if any job fails
+        required: false
+        default: true
+        type: boolean
       submodules:
         description: Whether to checkout submodules
         required: false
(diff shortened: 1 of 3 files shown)


astrofrog (approved):
Looks great! I agree with the defaults

Cadair (approved):

astrofrog:
@ConorMacBride @Cadair after using this I think it would be nice to have fast fail be false by default for wheels too - it would be good to see all errors when there are issues - it's more that we just don't want to *upload* if there are any issues.

PR #36, opened 2022-04-03
Upload to Anaconda.org

### LINKED ISSUE #8: Add documentation on publishing to alternative artifact feeds
The Azure Pipelines publishing template supported publishing to [Azure Artifacts](https://openastronomy-azure-pipelines.readthedocs.io/en/latest/publish.html#publishing-to-azure-artifacts). We should investigate how to do this here (either with Azure Artifacts or similar) and document. (Document how to configure the authentication etc.)

ConorMacBride:
This adds support for uploading to Anaconda.org in both of the publishing workflows. Matplotlib are now doing something similar for uploading nightly builds to Anaconda.org: https://github.com/matplotlib/matplotlib/blob/main/.github/workflows/nightlies.yml 
Although they are uploading the latest artifact instead of building new wheels especially for uploading.

Closes #8

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -41,6 +41,16 @@ on:
         required: false
         default: ''
         type: string
+      upload_to_anaconda:
+        description: A condition specifying whether to upload to Anaconda.org
+        required: false
+        default: false
+        type: boolean
+      anaconda_user:
+        description: Anaconda.org user or organisation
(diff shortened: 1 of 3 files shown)


astrofrog (approved):
I agree that it's a pain that this requires special treatment - though at least it works, and it's somewhat beyond our control? 😕

PR #42, opened 2022-04-11
Pure/non-pure package based on current workflow

### LINKED ISSUE #41: Pure/non-pure Python package testing
Now that https://github.com/OpenAstronomy/build-python-dist/pull/12 is merged, [`test_publish_pure_python.yml`](https://github.com/OpenAstronomy/github-actions-workflows/blob/main/.github/workflows/test_publish_pure_python.yml) is now failing because this repo is using a non-pure Python package to test the pure Python workflow. However, if we change it to a pure Python package cibuildwheel will complain instead.

As both of the publish workflows work by checking out and building/testing the current ref of the current repository, we can't tell it to checkout an external repo during the test. And we can't include two packages in the same repo as they both need different `setup.py` files.

Can you use the name of current workflow (available via the `GITHUB_WORKFLOW` env var) to include/exclude the compiled code inside `setup.py` and `__init__.py` etc.?

An alternative solution would be: 
1. Create two new repositories: a sample pure Python package, and a sample non-pure Python package
2. Add two new inputs to each of the workflows, `repository` and `ref`, which get passed to the checkout action
3. Update the tests in this repo to point them to the `main` branch of the correct external repository

Can you think of a better way?

ConorMacBride:
If the workflow named ".github/workflows/test_publish.yml" is running, the package will be non-pure. Otherwise, it will be pure Python.

Fixes #41

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,3 +2,6 @@
 requires = ["setuptools>=43.0.0",
             "wheel"]
 build-backend = 'setuptools.build_meta'
+
+[tool.cibuildwheel.linux]
+environment-pass = ["GITHUB_WORKFLOW"]
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,11 @@
 import os
 
-from setuptools import Extension, setup
+if os.getenv("GITHUB_WORKFLOW") == ".github/workflows/test_publish.yml":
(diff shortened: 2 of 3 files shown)


Cadair (approved):
Smooth

PR #46, opened 2022-05-18
Add support for caching

### LINKED ISSUE #44: Add ability to cache directories
It would be nice to have the option to cache a directory in at least the docs and tox workflows, using the actions/cache reusable action; https://github.com/actions/cache

ConorMacBride:
This integrates [`actions/cache`](https://github.com/actions/cache) into the tox workflow.

Closes #44


README.md
```diff
@@ -150,6 +150,21 @@ Whether to setup a headless display.
...
+
+#### cache-key
+An explicit key for restoring and saving the cache.
+Passed to [`actions/cache`](https://github.com/actions/cache) `key` input.
+Optional.
+
+#### cache-restore-keys
+An ordered list of keys to use for restoring the cache if no cache hit occurred for key.
+Passed to [`actions/cache`](https://github.com/actions/cache) `restore-keys` input.
+Optional.
```
Cadair, line 166: I think it would be useful (certainly for me) to have some examples of formulating the cache-key etc here. The example is the tests is a nice one for inter-job caches, but an example of a global pip cache or something would also be nice.

Cadair:
Looks nice!

README.md
ConorMacBride: I added two examples. One is a global cache based on the hash of a lock file, and the other is an inter-job cache where data are only downloaded once per run. I didn't give a specific pip caching example as I would need to be more sure about which specific directories need to be cached.

ConorMacBride:
This should be ready now!

Cadair (approved):

PR #55, opened 2022-05-24
Various documentation improvements

### LINKED ISSUE #47: Remove `hashFiles` function from example in caching docs
The repository is not checked out (and cannot be) at this stage of the example so the call to `hashFiles` in the cache key needs to be changed to something else: https://github.com/OpenAstronomy/github-actions-workflows#cache-path

A workaround would be for a user to run an initial job before calling the reusable workflow job. This job can checkout the repository, call `hashFiles` and set the hash as an output which can be used in the `cache-key` of subsequent jobs.

### LINKED ISSUE #50: Clarify required scope of anaconda token
The anaconda token needs the following scope:

Allow write access to the API site

We should mention this in the README

ConorMacBride:
Closes #47 
Closes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,20 +155,6 @@ A list of files, directories, and wildcard patterns to cache and restore.
 Passed to [`actions/cache`](https://github.com/actions/cache) `path` input.
 Optional.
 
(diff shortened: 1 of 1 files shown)


Cadair (approved):
Looks good.

Is it time we built a proper docs site?

PR #56, opened 2022-05-26
Add option to set environment variables when publishing

ConorMacBride:
Not sure about this. I think it would be neater as an independent JavaScript action. I couldn't find any existing actions that would accept a string as input.

Alternatively, if people give the env variables in a format which can be written to the `GITHUB_ENV` file directly, we wouldn't need to use Python. That is, give:
```yaml
env: |
  VAR1=test
  VAR2<<EOF
  first line
  seconds line
  EOF
  VAR3=testing
```
instead of 
```yaml
env: |
  VAR1: test
  VAR2: |
    first line
    seconds line
  VAR3: testing
```

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -26,6 +26,11 @@ on:
         required: false
         default: ''
         type: string
+      env:
+        description: A map of environment variables to be available when building and testing
+        required: false
+        default: ''
+        type: string
       libraries:
         description: Packages needed to build the source distribution for testing (installed using apt)
         required: false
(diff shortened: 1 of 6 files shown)


Cadair:
I wouldn't be adverse to telling people to give us the text of a thing that bash is going to source? (is that what's happening here?)

ConorMacBride:
To pass env variables to later steps in the same job, you [write them to the file defined in the `GITHUB_ENV` env variable](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable). I assume this means that each step is run within a separate shell, and therefore later steps wouldn't have access to the variables exported while sourcing the script the user provided.

Although it would work if people know to set env variables using the GitHub Actions way I linked above and not `export var=value`. E.g.

```yaml
before-build: |
  echo "MYVAR=hello world" >> $GITHUB_ENV
```
What do you think?

Cadair:
Does it have to be a file? I would rather have it configured in the yaml.

I could throw together a js action which converted yaml into the env file?

ConorMacBride:
A javascript action to do that would be ideal. You could even extend it to only set variables under particular conditions. E.g.:

```yaml
env: |
  if ${{ endsWith(github.ref, '.dev') }}:
    VAR1: test
  VAR2: |
    first line
    seconds line
  VAR3: testing
```

Cadair:
Is there a way to resolve the template variables like that easily?

ConorMacBride:
I think GitHub should resolve it before passing as input. So the action would receive:

```yaml
env: |
  if true:
    VAR1: test
  VAR2: |
    first line
    seconds line
  VAR3: testing
```

Cadair:
ooh

Cadair:
I am going to experiment on astropy.

Cadair:
Looks good to me!

nabobalis:
That also looks good to me.

Cadair:
If you can find the time to fix the conflicts then feel free to merge it.

Cadair (approved):

lpsinger:
I happened upon this PR while reviewing a dependabot update for one of the project that I co-maintain (https://github.com/astropy/astropy-healpix/pull/228). The base64-encoded Python scripts look fishy because they resemble code obfuscation like one might use to hide a malicious binary blob in open-source code. While their payloads appear to be innocuous, I think that they create the risk of someone sneaking malicious code into the publish workflow in the future. Can you do what you are trying to do without base64 encoding the Python scripts?

PR #57, opened 2022-06-01
Add `timeout-minutes` input

ConorMacBride:
Now [`timeout-minutes`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) can be specified for:

- Each tox environment in `tox.yml` (a global default value with a per tox environment override)
- Each sdist and wheel publish job in `publish.yml` (a global default **with no per build override** as this would need more work to implement)
- The whole `publish_pure_python.yml` workflow

The default value is `360` minutes, which is the GitHub default for the job level [`timeout-minutes`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes).


astrofrog (approved):
Looks good, thanks! Feel free to merge if this is ready

PR #61, opened 2022-06-17
Use conda to install `anaconda-client`

ConorMacBride:
Fixes a bug with uploading to Anaconda.org seen at https://github.com/astropy/astropy/issues/13348. `anaconda-client` is intended to be installed by conda (not pip, as some of its dependencies are not on PyPI). This PR switches the install method for this package from pip to conda.

- [x] Delete TMP commit before merging

cc @pllim

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -183,11 +183,13 @@ jobs:
           user: __token__
           password: ${{ secrets.pypi_token }}
           repository_url: ${{ inputs.repository_url }}
+      - uses: conda-incubator/setup-miniconda@v2
+        if: ${{ inputs.upload_to_anaconda }}
       - name: Upload to Anaconda.org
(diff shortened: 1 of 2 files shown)


ConorMacBride:
By temporarily removing the Anaconda upload condition, `anaconda-client` is installing successfully with conda and the `anaconda` command is available: https://github.com/OpenAstronomy/github-actions-workflows/runs/6941673520?check_suite_focus=true#step:9:298

This should be ready to merge now. Then I'll do a bugfix release.

pllim:
Thanks! So, once you release this, astropy wheel job would automatically work again?

ConorMacBride:
Yes, although I think it might still fail due to https://github.com/astropy/astropy/issues/13197#issuecomment-1146674168

pllim (approved):
Thanks!

PR #66, opened 2022-08-02
Use publish-wheels-anaconda

astrofrog:
This switches to the new Action I wrote to upload the files to anaconda and optionally remove old versions (to prevent build-up of too many developer versions).

A few questions/open issues before this can be merged:

* Are we happy with the name of the action or can we think of a better name? (I'm not attached to the current one)
* Is there a way to avoid having to define anaconda_package?
* Currently the action hard-codes dist/* but we could potentially allow that to be customized
* Would it make sense to test the anaconda upload here somehow? In the action we do test uploading to the anaconda.org for real using a test user so maybe we could do the same here?

I don't have much time to work on this in the near future so if someone wants to take the charge on this please feel free to!

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -201,7 +201,7 @@ jobs:
           password: ${{ secrets.pypi_token }}
           repository_url: ${{ inputs.repository_url }}
       - uses: OpenAstronomy/publish-wheels-anaconda@main
-        if: ${{ needs.targets.outputs.upload_to_anaconda == 'true' }}
+        if: ${{ inputs.upload_to_anaconda }}
         with:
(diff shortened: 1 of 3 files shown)


Cadair (approved):
Looks sane to me if it works then :+1:

ConorMacBride (approved):
> Are we happy with the name of the action or can we think of a better name? (I'm not attached to the current one)

I don't mind. I think it's fine.

> Would it make sense to test the anaconda upload here somehow? In the action we do test uploading to the anaconda.org for real using a test user so maybe we could do the same here?

Yeah, it should be easy to use the same test credentials here.

---

If you're happy with the name, I think this is ready to merge 👍

PR #80, opened 2022-10-04
Do not append OS name to a job name explicitly given

Cadair:
Opening this on an upstream branch for easy testing on astropy.


tools/tox_matrix.py
```diff
@@ -126,7 +126,7 @@ def get_matrix_item(env, global_libraries, global_string_parameters,
         item["python_version"] = env.get("default_python") or default_python
 
     # set name
-    item["name"] = env.get("name") or item["toxenv"]
+    item["name"] = env.get("name") or f'{item["toxenv"]} ({item.os})'
```
ConorMacBride, line 129:
```suggestion
    item["name"] = env.get("name") or f'{item["toxenv"]} ({item["os"]})'
```

Cadair: :facepalm:

ConorMacBride (approved):

Issue #64, opened 2022-07-21
Use publish-wheels-anaconda


astrofrog:
I've now got a working version of https://github.com/OpenAstronomy/publish-wheels-anaconda which includes removing old files - we could perhaps use this in the workflows here?

PR #84, opened 2022-10-15
Fix deprecated `set-output` command

### LINKED ISSUE #83: `set-output` command is deprecated
In ASDF, I have noticed the following warning being raised by GitHub concerning the CI that is based on the OpenAstronomy workflows:

```
core / Load tox environments
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

See https://github.com/asdf-format/asdf/actions/runs/3245945344 for details of the run.

According to https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/, it appears that `set-output`, which is used during the "Load tox environments" step is now being depreciated.

ConorMacBride:
Converts to new method: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

Fixes #83


Cadair (approved):

Cadair:
This is probably worthy of a new release? Shall we get this and #64 in and then tag?

ConorMacBride:
Sounds good. I'll do a release now

PR #98, opened 2022-12-24
Allow passing Python version in pure python publish workflow

wtbarnes:
This allows passing in a Python version when using the pure Python publish workflow. I think this should propagate to the build-python-dist workflow.

### DIFF
--- a/.github/workflows/publish_pure_python.yml
+++ b/.github/workflows/publish_pure_python.yml
@@ -91,6 +91,7 @@ jobs:
         with:
           test_extras: ${{ inputs.test_extras }}
           test_command: ${{ inputs.test_command }}
+          python-version: ${{ inputs.python-version }}
           pure_python_wheel: true
       - id: set-upload
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_pure_python.yml
```diff
@@ -91,6 +91,7 @@ jobs:
         with:
           test_extras: ${{ inputs.test_extras }}
           test_command: ${{ inputs.test_command }}
+          python-version: ${{ inputs.python-version }}
```
Cadair, line 94: Does this not need a default specified somewhere to maintain backward compatibility?

wtbarnes: Yes. I think the default value is '3.x'. I didn't implement that here because I wasn't sure how the templating logic worked.

wtbarnes: Ok I *think* I've now implemented the argument correctly and have just used the same defaults as the `build-python-dist` workflow.

ConorMacBride (commented):
This should work. It just needs the option listed in the documentation: https://github.com/OpenAstronomy/github-actions-workflows#inputs-2

wtbarnes:
Ok I've updated the docs with a note about `python-version`. Some of the 'test-libaries' checks are failing or reasons I do not understand.

ConorMacBride (approved):

Issue #62, opened 2022-06-21
Add support for pip dependency caching


Cadair:
The built in pip cache support to `setup-python` looks pretty easy to use: https://github.com/actions/setup-python/#caching-packages-dependencies

We could probably enable it with a single variable giving the value for `cache-dependency-path`?

v-parmar:
Hi, I can Do it using Tox.

PR #102, opened 2023-01-02
Add support for Python package caching

dstansby:
Fixes https://github.com/OpenAstronomy/github-actions-workflows/issues/62

### DIFF
--- a/.github/workflows/test_tox.yml
+++ b/.github/workflows/test_tox.yml
@@ -130,3 +130,10 @@ jobs:
           cache-path: |
             a/
           cache-key: cache-${{ github.run_id }}
+
+  test_no_python_cache:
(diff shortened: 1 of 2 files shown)


Cadair:
Having had another look at the description on the setup-python action, I think we are going to need to tell it what files to use to control the cache?

> Input cache-dependency-path is used for cases when multiple dependency files are used, they are located in different subdirectories or different files for the hash that want to be used.

Although i don't know what files sunpy would use to control the cache as it would need updating regularly to catch the newest files and we don't have a pins file?

dstansby:
Yeah, I wonder if it's easiest for users of this action to do this manually using the caching arguments already present? At least this PR isn't a general enough approach, so I'll close.

PR #105, opened 2023-01-31
Fix coverage upload for astropy

pllim:
astropy coverage upload keeps failing. This patch shows you the settings we had before. You don't need a token for public repo. And if you cannot accept this patch, please suggest an alternative. Otherwise I am not sure what to do.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -181,4 +181,5 @@ jobs:
         if: ${{ contains(matrix.coverage, 'codecov') && matrix.pytest == 'true' }}
         uses: codecov/codecov-action@v3
         with:
-          token: ${{ secrets.CODECOV_TOKEN }}
+          file: ./coverage.xml
+          verbose: true
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -181,5 +181,6 @@ jobs:
         if: ${{ contains(matrix.coverage, 'codecov') && matrix.pytest == 'true' }}
         uses: codecov/codecov-action@v3
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           file: ./coverage.xml
           verbose: true


pllim:
I don't think this error is related?

`libraries: failed with dot is not allowed, use allowlist_externals to allow it`

pllim:
@ConorMacBride , please advise. Thanks!

.github/workflows/tox.yml
```diff
@@ -181,4 +181,5 @@ jobs:
         if: ${{ contains(matrix.coverage, 'codecov') && matrix.pytest == 'true' }}
         uses: codecov/codecov-action@v3
         with:
-          token: ${{ secrets.CODECOV_TOKEN }}
```
ConorMacBride, line 184: Removing this should not make any difference because whether the default `secrets.CODECOV_TOKEN` is null or an empty string, it is interpreted the same as not being set at all: https://github.com/codecov/codecov-action/blob/e0fbd592d323cb2991fb586fdd260734fcb41fcb/src/buildExec.ts#L78-L80

.github/workflows/tox.yml
```diff
@@ -181,4 +181,5 @@ jobs:
...
+          file: ./coverage.xml
```
ConorMacBride, line 184: Adding this line should be okay, however, if it's needed, it is a bug in codecov. 

We're always writing to `$GITHUB_WORKSPACE/coverage.xml` even if tox is running inside an isolated directory. This location should always be searched by default. https://github.com/OpenAstronomy/github-actions-workflows/blob/65452d98073d10f22caedc746d6e9a27e1a75320/tools/tox_matrix.py#L138

pllim: I am really not sure what is happening upstream. Just following some ideas from https://github.com/codecov/codecov-action/issues/598

pllim:
Maybe not needed.

PR #107, opened 2023-02-14
Allow env vars to be passed to tox workflow

ConorMacBride:
Similar to #56 but extended to the tox workflow.


pllim:
Is there a reason why this isn't merged? Could use this for https://github.com/astropy/astropy/pull/14401 or https://github.com/astropy/astropy/issues/14161

ConorMacBride:
Glad it can be of use! Was just waiting for some reviews before committing to the API. The publishing workflows call the input `env` but I called it `setenv` in this PR so it isn't confused with the existing `envs` input in the tox workflow.

pllim:
As long as that distinction is clearly documented, I am neutral on the name. Thanks!

pllim:
Thanks for merging! Any plans to do a release so we can pick this up downstream under `v1`?

ConorMacBride:
Yeah, could do. I think there's been quite a few features added. I'll do a release later today.

PR #114, opened 2023-03-15
add Python version to name

zacharyburnett:
this might cause some issues with required checks in repository branch protection settings (which look for an exact name match) so perhaps should be cause for a bump to `@v2`

### DIFF
--- a/tools/tox_matrix.py
+++ b/tools/tox_matrix.py
@@ -130,7 +130,7 @@ def get_matrix_item(env, global_libraries, global_string_parameters,
         item["python_version"] = env.get("default_python") or default_python
 
     # set name
-    item["name"] = env.get("name") or f'{item["toxenv"]} ({item["os"]})'
+    item["name"] = env.get("name") or f'{item["toxenv"]} (Python {item["python_version"]} on {item["os"]})'
 
     # set pytest_flag
     item["pytest_flag"] = ""


ConorMacBride:
Yeah, this would require changes to required checks. I'd want to think about this more before bumping the major version for an (arguably) cosmetic change. I think most users wouldn't need the Python version displayed here because it's either [already part of the tox environment name](https://github.com/sunpy/sunpy/actions/runs/4419454596/jobs/7747974964) or they [manually set how their tox environments are displayed by this workflow](https://github.com/astropy/astropy/blob/c0a24c1dc957a3b565294213f435fefb2ec99714/.github/workflows/ci_workflows.yml#L88). And for other jobs, such as linting, it is less important to know the Python version.

In tox you can [prefix environment names with e.g. `py310-`](https://tox.wiki/en/latest/user_guide.html#test-environments) to specify the Python version (for the top level `[testenv]` section tests). The links above show examples of this. Are you intending to use this for https://github.com/spacetelescope/jwst/pull/7498? You could configure it as:
```yaml
      envs: |
        - linux: py38-test-oldestdeps-xdist-cov
        - linux: py38-test-xdist
        - linux: py39-test-xdist
        - linux: py39-test-sdpdeps-xdist
        - linux: py310-test-xdist
        - linux: py310-test-opencv-xdist
        - linux: py311-test-xdist
        - macos: py311-test-xdist
        - linux: test-xdist-cov
          coverage: codecov
```
This workflow will then [determine the python version to setup](https://github.com/OpenAstronomy/github-actions-workflows/blob/62c10bd5782a5a0e196a6c713c6db02d2392f234/tools/tox_matrix.py#L128) based on the environment name. That should work without any modifications to [`tox.ini`](https://github.com/spacetelescope/jwst/blob/master/tox.ini).

zacharyburnett:
ah, that is a good point, thank you! I will close this PR then

Issue #81, opened 2022-10-10
Upload coverage to codecov, even if some tests fail


dstansby:
In sunpy we often have remote data tests failing because of remote data provider issues. In this case we often merge PRs with failing tests, but want the code coverage to be measured on the remote data runs that have some failing tests. It would therefore be good to uplaod code coverage even if some tests fail.

Cadair:
It looks like this is something that you can configure directly with codecov: https://docs.codecov.com/docs/common-recipe-list#see-coverage-on-ci-test-failures

ConorMacBride:
The codecov config doesn't control whether results are uploaded, just how the report is presented. We would need to add an ["if: always"](https://docs.github.com/en/actions/learn-github-actions/expressions#always) to the upload step, but that may not be wanted in most cases.

Cadair:
Ah, I see.

PR #122, opened 2023-04-19
Upload to codecov, even when some tests fail

dstansby:
Fixes https://github.com/OpenAstronomy/github-actions-workflows/issues/81. I think it's safe to always run the codecov step, no matter what happens in the tox step. If no coverage file is generated, codecov just won't upload or report anything.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -194,7 +194,8 @@ jobs:
         shell: sh
 
       - name: Upload to Codecov
-        if: ${{ contains(matrix.coverage, 'codecov') && matrix.pytest == 'true' }}
+        # Even if tox fails, upload coverage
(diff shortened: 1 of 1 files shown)


Cadair (approved):
I am fine with it, and definitely pro this for sunpy.

ConorMacBride (changes requested):
Setting this will mean that the job (tox environment) will [always be reported as successful](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error). We would need to do the following instead:

```yaml
- name: Upload to Codecov
  if: ${{ (success() || failure()) && contains(matrix.coverage, 'codecov') && matrix.pytest == 'true' }}
```

I think it's fine to always upload the coverage. If people don't like it, they can just pass `--no-cov-on-fail` to pytest. I imagine for most projects, if the tests fail, the coverage status is meaningless anyway.

ConorMacBride (approved):

PR #123, opened 2023-04-21
use micromamba

### LINKED ISSUE #197: Fix conda support on MacOS
As seen in the CI, conda doesn't work anymore on MacOS

One solution might be to try conda-incubator/setup-miniconda instead of the current setup-conda action.

zacharyburnett:
fixes #197 I think?

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -189,7 +189,9 @@ jobs:
         with:
           environment-name: test
           create-args: >-
+            conda
             python=${{ matrix.python_version }}
+            tox
           cache-environment: true
           cache-downloads: true
 
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -209,7 +209,6 @@ jobs:
           echo $SET_ENV_SCRIPT | base64 --decode > set_env.py
           python set_env.py "${{ matrix.setenv }}"
           rm set_env.py
-        shell: sh
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/tox.yml
```diff
@@ -166,10 +166,15 @@ jobs:
...
-        uses: s-weigand/setup-conda@v1
+        uses: mamba-org/provision-with-micromamba@v15
         with:
-          update-conda: true
-          python-version: ${{ matrix.python_version }}
+          environment-file: false
+          environment-name: 'test'
+          extra-specs: |
+            python=${{ matrix.python_version }}
+          channels: defaults
```
ConorMacBride, line 175: Should we change this to `conda-forge`? I guess this will use the Anaconda `defaults` channel for installing Python, but people can then specify `conda-forge` in their tox-conda config?

ConorMacBride (changes requested):
Looks good. Just one question about the conda channel. Also, I think the changes to `test_tox.yml` and `tox.ini` should be reverted. I added tests with the `con_da` tox environment is for testing that conda can be configured via an Actions input, and not just through the environment name. (The tests for this project are quite confusing!)

.github/workflows/tox.yml
zacharyburnett: Good point; it might just be better to use `conda-forge` by default.

zacharyburnett:
> I think the `assert len(weights) == expected_node_count` was an error in pip and was fixed in pip v22.0.3, released over a year ago. Since the pip version is so old, I wonder if `python` is the [system python](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#package-management) rather than the one installed by micromamba?
> 
> Maybe we should be avoiding `pip` when using `conda`? We could install tox and the tox-deps by passing them into `create-args` instead?
> 
> Also, I [this silent error](https://github.com/OpenAstronomy/github-actions-workflows/actions/runs/4823459898/jobs/8592019267?pr=123#step:10:9) looks like we might need to update the `conda` executable to `micromamba` in `tox.ini`.

@ConorMacBride sorry it took me so long to get back to this! The tests appear to be passing now

zacharyburnett:
not to rush anyone; this is ready to merge and fixes the failing tests

ConorMacBride (approved):
Thanks @zacharyburnett! Sorry for the delay in reviewing

PR #127, opened 2023-04-23
Only set CIBW testing env vars if workflow input is given

### LINKED ISSUE #126: Not possible to put cibuildwheel test or extras config in pyproject.toml
Because the publish workflow specifies the follwing environment variables:
https://github.com/OpenAstronomy/github-actions-workflows/blob/bb12d5420b890a2102eea80fbfbda86e626baec4/.github/workflows/publish.yml#L167-L168

it's not possible to instead configure these cibuildwheel options in a `pyproject.toml` file, because not setting these inptus sets the environment variables to empty strings, which overrides `pyproject.toml` configuration.

Possible solutions:
- Only set these environment variables if the inputs are given to the action
- Remove these variables, and instead tell users to use `pyproject.toml` to configure cibuildwheel. This would be breaking, but I think with a major version bump of the action fine.

ConorMacBride:
Check if the `test_extras` or `test_command` workflow inputs are given, and only set the CIBW environment variables if they are. Therefore cibuildwheel can fall back to configuration in 
`pyproject.toml`.

Closes #126

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -143,6 +143,18 @@ jobs:
         uses: docker/setup-qemu-action@v2
         with:
           platforms: all
+      - name: Configure cibuildwheel
+        shell: bash
+        run: |
+          if [ -n "${{ inputs.test_extras }}" ];
+          then
+            echo "CIBW_TEST_EXTRAS=${{ inputs.test_extras }}" >> $GITHUB_ENV
+          fi
+          if [ -n "${{ inputs.test_command }}" ];
+          then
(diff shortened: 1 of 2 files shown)


ConorMacBride:
Most projects will still need to set the `test_extras` and `test_command` inputs for the sdist build anyway. I wonder if people would expect their `pyproject.toml` cibuildwheel config to override the workflow inputs? In any case, I think checking `pyproject.toml` configuration would overcomplicate this workflow, and this fix is good enough.

Cadair (approved):

PR #130, opened 2023-04-27
display test results in GitHub step summary

zacharyburnett:
uses the [`test-summary` action](https://github.com/marketplace/actions/testforest-dashboard) to send test results to the `$GITHUB_STEP_SUMMARY` feature


.github/workflows/tox.yml
```diff
@@ -190,9 +195,14 @@ jobs:
         run: python -m pip install --upgrade tox ${{ matrix.toxdeps }}
         shell: sh
 
-      - run: python -m tox -e ${{ matrix.toxenv }} ${{ matrix.toxargs }} -- ${{ matrix.pytest_flag }} ${{ matrix.posargs }}
+      - run: python -m tox -e ${{ matrix.toxenv }} ${{ matrix.toxargs }} -- ${{ matrix.pytest_flag }} ${{ matrix.posargs }} ${{ fromJSON('["--junitxml results.xml", ""]')[matrix.pytest_results_summary == 'true' && matrix.pytest == 'true'] }}
```
zacharyburnett, line 198: @ConorMacBride is there a mechanism to run `--junitxml results.xml` and the subsequent action only when `tox` is running `pytest`? Or would that need to be included in the `toxenv`?

ConorMacBride: Yeah, that should be possible. We do that for the coverage report here:
https://github.com/OpenAstronomy/github-actions-workflows/blob/a7918d02a33a34288dd0182170b39c74a71ba81e/tools/tox_matrix.py#L135-L140
The `pytest_results_summary` input will need to be integrated into `tools/tox_matrix.py` for it to be included in the job matrix. Will need to be included in the "string parameter" dictionaries and also included as a `@click.option` decorator.

zacharyburnett: Great! Should I manually serialize the script into the workflow after making the change?

ConorMacBride: If you have pre-commit installed, you can just do `pre-commit run --all-files` and it'll update the serialization for you.

tools/tox_matrix.py
```diff
@@ -138,6 +140,8 @@ def get_matrix_item(env, global_libraries, global_string_parameters,
         sep = r"\\" if platform == "windows" else "/"
         item["pytest_flag"] += (
             rf"--cov-report=xml:${{GITHUB_WORKSPACE}}{sep}coverage.xml ")
+    if item["pytest"] and item["pytest-results-summary"]:
```
ConorMacBride, line 143: These are strings
```suggestion
    if item["pytest"] == "true" and item["pytest-results-summary"] == "true":
```

tools/tox_matrix.py
```diff
@@ -138,6 +140,8 @@ def get_matrix_item(env, global_libraries, global_string_parameters,
         sep = r"\\" if platform == "windows" else "/"
         item["pytest_flag"] += (
             rf"--cov-report=xml:${{GITHUB_WORKSPACE}}{sep}coverage.xml ")
+    if item["pytest"] == "true" and item["pytest-results-summary"] == "true":
+        item["pytest_flag"] += "--junitxml results.xml"
```
ConorMacBride, line 144: Depending on how someone configures their `tox.ini`, I think it might be possible for this file to created in a temp directory, for example. So maybe it would be safer to configure it similar to the cov-report path (`${{GITHUB_WORKSPACE}}{sep}results.xml`)?

.github/workflows/tox.yml
```diff
@@ -32,6 +32,11 @@ on:
         required: false
         default: true
         type: boolean
+      pytest_results_summary:
```
ConorMacBride, line 35:
```suggestion
      pytest-results-summary:
```

.github/workflows/test_tox.yml
```diff
@@ -151,3 +151,10 @@ jobs:
           cache-path: |
             a/
           cache-key: cache-${{ github.run_id }}
+
+  test_pytest_args:
+    uses: ./.github/workflows/tox.yml
+    with:
+      envs: |
+        - linux: py310-linux
+          pytest_results_summary: true
```
ConorMacBride, line 160:
```suggestion
          pytest-results-summary: true
```

tools/tox_matrix.py
```diff
@@ -138,6 +140,8 @@ def get_matrix_item(env, global_libraries, global_string_parameters,
         sep = r"\\" if platform == "windows" else "/"
```
ConorMacBride, line 140: This will need to be before the if block now

zacharyburnett: thanks! Sorry for all the changes 😅

.github/workflows/tox.yml
```diff
@@ -193,6 +198,11 @@ jobs:
       - run: python -m tox -e ${{ matrix.toxenv }} ${{ matrix.toxargs }} -- ${{ matrix.pytest_flag }} ${{ matrix.posargs }}
         shell: sh
 
+      - if: ${{ (success() || failure()) && matrix.pytest_results_summary == 'true' && matrix.pytest == 'true' }}
```
ConorMacBride, line 201:
```suggestion
      - if: ${{ (success() || failure()) && matrix.pytest-results-summary == 'true' && matrix.pytest == 'true' }}
```

zacharyburnett:
It looks like the summary works, at least to determine a lack of tests:
image

ConorMacBride:
Can you also list this option in the readme (just under https://github.com/OpenAstronomy/github-actions-workflows#pytest). Then I think this is ready to be merged! Thanks :)

zacharyburnett:
README updated: https://github.com/zacharyburnett/github-actions-workflows/tree/step_summary#pytest-results-summary

ConorMacBride (approved):
Brilliant, looking forward to using this! Thank you very much 🚀

PR #131, opened 2023-04-27
ignore changes to other workflows when dispatching tests

zacharyburnett:
only run `test_tox.yml`, `test_publish.yml`, and `test_publish_pure_python.yml` when their respective workflows change


ConorMacBride (approved):
Excellent idea! Thanks 🚀

PR #137, opened 2023-05-08
fix brackets escape for github workflow inline expression

zacharyburnett:
The output path for the test results XML should be `${{GITHUB_WORKSPACE}}/results.xml`, but instead it is `${GITHUB_WORKSPACE}/results.xml`; this is due to the way that brackets are escaped in f strings 

image


ConorMacBride:
`${GITHUB_WORKSPACE}/results.xml` should be expanded by the shell. Alternatively, `${{ github.workspace }}/results.xml` would make GitHub Actions expand it before running it. [This workflow run](https://github.com/sunpy/sunpy/actions/runs/4926325416/jobs/8803209092?pr=7000#step:12:36) wrote the coverage results to the correct location even with `${GITHUB_WORKSPACE}`. Have you come across a case where this does not expand the path as expected?

PR #142, opened 2023-06-30
Allow checkout ref to be customized

astrofrog:
I am working on a package (astropy-iers-data) for which I would like to have completely automated releases - this involves having a workflow where I update some files automatically, push a commit to the repository, push a tag to the repository, and then run the publish workflow here. However, by default, the checkout step doesn't see the latest commit I created in the workflow, so I need to be able to pass a checkout ref to the checkout step. I may have missed a better way to do this though?

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -86,6 +86,11 @@ on:
         required: false
         default: true
         type: boolean
+      checkout_ref:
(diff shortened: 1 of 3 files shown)


Cadair (approved):
Looks fine to me. If it's needed then why not?

PR #143, opened 2023-07-07
DOC: test_command fails when double quotes exist

pllim:
`test_command` fails when double quotes exist, see https://github.com/astropy/astropy/pull/15040 

cc @saimn @dhomeier

### DIFF
--- a/docs/source/publish.rst
+++ b/docs/source/publish.rst
@@ -82,6 +82,8 @@ The command to run to test the package. Will be run in a temporary
 directory. Default is no testing.
 If not set, cibuildwheel will use any ``test-command`` configured in ``pyproject.toml``.
 
+If you need quotes in the command, please use single quotes, not double quotes.
+
 env
 ^^^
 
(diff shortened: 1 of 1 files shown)


docs/source/publish.rst
```diff
@@ -82,6 +82,8 @@ The command to run to test the package. Will be run in a temporary
 directory. Default is no testing.
 If not set, cibuildwheel will use any ``test-command`` configured in ``pyproject.toml``.
 
+If you need quotes in the command, please use single quotes, not double quotes.
```
dhomeier, line 85:
```suggestion
This input will be treated as a string in double quotes similar to an environment variable
by some workflows; therefore if you need quotes within the command, please use
single quotes, not double quotes.
```

pllim: Thanks!

ConorMacBride:
Thanks for highlighting this bug! Since it silently fails and lots of projects are likely to be including quotes in their test commands (e.g. sunpy also, but instead of not testing it fell back to the identical cibuildwheel test command defined in sunpy's `pyproject.toml`), I think it would be better to update the script to allow double quotes. I've opened #144 to fix the issue. Please let me know what you think of that PR.

Cadair:
Closing as superseded by #144

PR #148, opened 2023-08-18
Add upload artifact step

ConorMacBride:
Added a step in the tox workflow for https://github.com/actions/upload-artifact. A path to save as an artifact can be passed to each tox environment. I've added a test to confirm it can write an artifact, and then a separate job to download the artifact and confirm a file exists.


astrofrog (approved):
This is great, thanks!

PR #162, opened 2023-09-20
Try to block artifact uploads when unnecessary

### LINKED ISSUE #159: BUG: publish workflow tries to upload wheel artifact generated by PR
See https://github.com/astropy/astropy/pull/14784#issuecomment-1722259319

WilliamJamieson:
Attempt to fix #159.

This prevents artifact uploads when the wheels will not be uploaded to PyPi or anaconda.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -190,6 +190,8 @@ jobs:
           CIBW_BUILD: ${{ matrix.CIBW_BUILD }}
           CIBW_ARCHS: ${{ matrix.CIBW_ARCHS }}
       - uses: actions/upload-artifact@v3
+        if: |
+          needs.targets.outputs.upload_to_pypi == 'true' || inputs.upload_to_anaconda
         with:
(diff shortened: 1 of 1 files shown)


Cadair (commented):
Thanks for the PR. Looks like there's also an instance of this for the sdist and maybe also in the pure python publish workflow?

WilliamJamieson:
> Thanks for the PR. Looks like there's also an instance of this for the sdist and maybe also in the pure python publish workflow?

Looks like there is one for sdist, but there is not one for the pure python one.

WilliamJamieson:
@pllim I think this solves the issue you reported.

pllim (approved):
LGTM. Thanks!

ConorMacBride (approved):
Thanks @WilliamJamieson! If anyone was relying on downloading the build artifacts we can add an option to enable it. Although, since artifacts were not uploaded for the pure workflow I think it's fine to call this a bug.

Issue #167, opened 2023-11-10
`sdist` not being uploaded along with the wheels.


WilliamJamieson:
This is related to astropy/astropy#15597.

## Background
The astropy 6.0rc1 release was triggered, which triggered the building and uploading of the astropy 6.0rc1 wheels to PyPi, see https://github.com/astropy/astropy/actions/runs/6817064464/job/18540287448. The issue is that the source distribution was not uploaded to PyPi along side all the pre-built wheels.

It looks like the `Run actions/upload-artifact@v3` step was successfully triggered by all of the wheel builds, see for example https://github.com/astropy/astropy/actions/runs/6817064464/job/18539760549. However, for the source distribution `Run actions/upload-artifact@v3` was not triggered, see https://github.com/astropy/astropy/actions/runs/6817064464/job/18539753723.

As far as I can tell, the source distribution upload would only be blocked by: https://github.com/OpenAstronomy/github-actions-workflows/blob/59a929bdf64d072b19c8fc758ae44dd768f6f3fe/.github/workflows/publish.yml#L235-L236

This appears to be identical to the wheel upload blocker: https://github.com/OpenAstronomy/github-actions-workflows/blob/59a929bdf64d072b19c8fc758ae44dd768f6f3fe/.github/workflows/publish.yml#L193-L194

Yet the former was blocked while the latter was not. Diving into the logs we see that for both the source and wheel builds the inputs both clearly have `upload_to_pypi: true` meaning that in both cases the upload should have occurred, see

- Source: https://github.com/astropy/astropy/actions/runs/6817064464/job/18539753723#step:1:84
- Wheel: https://github.com/astropy/astropy/actions/runs/6817064464/job/18539760549#step:1:84

So I do not understand why one uploads while the other does not. There are no failures in any of the steps proceeding the failure.

WilliamJamieson:
Tagging @ConorMacBride and @Cadair for assistance.

astrofrog:
I've been investigating this but I also don't understand what is going on. Will keep trying, but in the meantime I think it should be safe to upload a locally built sdist.

WilliamJamieson:
The stcal package experienced the same issue (source distribution did not get uploaded to PyPi) today when it made a release. See the https://github.com/spacetelescope/stcal/actions/runs/6881432549/job/18717759512

astrofrog:
Ok I'll try and investigate some more

astrofrog:
[...]

PR #169, opened 2023-11-16
Make sure build_sdist requires targets as it is needed for the artifact upload condition

astrofrog:
This fixes https://github.com/OpenAstronomy/github-actions-workflows/issues/167

The issue was that the artifact upload condition in the sdist job is:

```
      - uses: actions/upload-artifact@v3
        if: |
          needs.targets.outputs.upload_to_pypi == 'true' || inputs.upload_to_anaconda
```

but the sdist job didn't technically require targets so ``needs.targets.outputs.upload_to_pypi`` wasn't set correctly. I don't quite understand why this wasn't an issue before, maybe GitHub got more strict about this, or maybe it is a race condition.

This also explains why astropy nightlies were working fine - the anaconda upload condition is separate and doesn't rely on ``needs.targets``.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -197,6 +197,7 @@ jobs:
 
   build_sdist:
     name: Build source distribution
+    needs: [targets]
     if: ${{ inputs.sdist }}
     runs-on: ${{ inputs.sdist-runs-on }}
     timeout-minutes: ${{ inputs.timeout-minutes }}


pllim:
Can we tag this in `v1` so downstream can pick it up without having to make changes? Thanks!

astrofrog:
Done already!

pllim:
Great. Thanks!

ConorMacBride:
Well spotted! The condition was only added in #162, so maybe there wasn't an astropy release since then?

PR #178, opened 2024-01-04
Upgrade `upload-artifact` and `download-artifact` actions from `v3` to `v4`

ConorMacBride:
The `v3` to `v4` upgrade introduces three breaking changes: https://github.com/actions/upload-artifact#breaking-changes

Two of them are not relevant as we don't support self-hosted runners, and we only ever save one artifact per job.

The other is that the artifacts are now immutable -- we cannot keep uploading wheels to the same artifact in each job. I've updated the workflow to write a separate artifact per build job and merge them in the upload job.

Supersedes #173 and #174.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -113,7 +113,7 @@ jobs:
       - run: python -m pip install PyYAML click
       - run: echo $LOAD_BUILD_TARGETS_SCRIPT | base64 --decode > load_build_targets.py
         env:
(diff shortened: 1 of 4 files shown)


pllim:
Thanks! Is this compatible with how astropy uses this workflow in https://github.com/astropy/astropy/blob/main/.github/workflows/publish.yml ? cc @astrofrog and @saimn

pllim:
For other downstream packages that uses this, will things just work or we have to change stuff too in our own repos?

pllim:
Given that some artifacts are uploaded as nightly wheels (say, to https://anaconda.org/astropy/astropy/files?type=pypi ), how will this affect the "old files pruning" script that @astrofrog previously implemented?

ConorMacBride:
Tested in this workflow run: https://github.com/OpenAstronomy/github-actions-workflows/actions/runs/7412831365/job/20170967258?pr=178#step:4:6

ConorMacBride:
> Thanks! Is this compatible with how astropy uses this workflow in https://github.com/astropy/astropy/blob/main/.github/workflows/publish.yml ? cc @astrofrog and @saimn

The astropy workflow looks to be compatible with the changes in this PR.

This PR should not introduce any breaking changes for any repo. The only new breakage would be workflows that are building the same target more than once which is likely a mistake – the astropy workflow doesn't have this issue because the upload to PyPI is disabled in the first job so only one artifact is saved for that duplicate build target.

ConorMacBride:
> Given that some artifacts are uploaded as nightly wheels (say, to https://anaconda.org/astropy/astropy/files?type=pypi ), how will this affect the "old files pruning" script that @astrofrog previously implemented?

This PR should not affect that. The publish workflow changes in this PR only affect how built wheels and sdists are passed from the build jobs to the upload job.

Cadair (approved):

pllim:
Can this be merged and a release made?

I dunno what is going on but the released upload workflow is terribly broken: https://github.com/astropy/astropy/actions/runs/7620855779/job/20756741326 (it keeps the very old wheel but deletes the new wheels that are supposed to be uploaded) 🔥

PR #186, opened 2024-02-13
Use `macos-14` M1 runner when building macOS ARM wheels

ConorMacBride:
This ~should~ does enable testing when building macOS ARM wheels.

- [x] Avoid hardcoding the M1 runner name


Cadair:
ooo! yay!

ConorMacBride:
`macos-latest` will point to `macos-14` anyway around [Apr – June 2024](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/) and then I guess `macos-latest` will always be an `arm64` runner. At that point we can just revert this PR.

Cadair:
Aren't we going to need the reverse then? Or can you run and test x86 mac builds on the arm platform?

Cadair (approved):

ConorMacBride:
You can build and test x86 on the Apple ARM systems. Apple have a translator for running x86 on ARM.

PR #195, opened 2024-04-23
Explicitly use macos-12 for intel builds

astrofrog:
Now that macos-latest is macos-14, we need to explicitly specify to use macos-12 for non-arm64 builds when building wheels. At some point we should probably switch around the logic to check for when the arch is x86_64 and set that to macos-12 and use macos-latest otherwise, but for now this should be sufficient.


astrofrog:
This is time-sensitive as I have some wheel jobs failing for intel wheels and I'll need to make sure it works for the upcoming astropy release.

PR #196, opened 2024-04-25
Fix macos image to use in tox jobs

astrofrog:
I am seeing failures in packages that have py38- and py39- tox environments being tested on Mac since those Python versions are not available on macos-14 which is now the macos-latest. So for <3.10, if the OS is macos-latest, we need to hard-code it to macos-12 which will be the last image to ever support these Python versions.


astrofrog:
Conda is broken, not sure why but not related to the changes here - I did see the conda issue in other repos, so it might be an upstream issue related to the macos-latest migration.

dhomeier:
Which is a bit strange to begin with, as miniconda still provides 3.8.19 and 3.9.19 on macOS 14.4.1 (which even has 3.9.6 as system Python).

Cadair:
Now that github have pulled macos-12 runners should we just revert this?

PR #199, opened 2024-05-08
Coverage: Upload as separate step

pllim:
Upload to Codecov as separate step so it can be restarted without re-running the whole test suite because sometimes Codecov rate limits.

We need this to fix https://github.com/astropy/astropy/issues/16379

This PR is tested downstream at https://github.com/astropy/astropy/pull/16419

Also see https://github.com/spacetelescope/jdaviz/pull/2865

You will see deprecation warning from https://github.com/actions/download-artifact/issues/283 but it won't fail the upload.

cc @ConorMacBride @bsipocz


pllim:
The test-conda OSX failure is unrelated (https://github.com/OpenAstronomy/github-actions-workflows/issues/197)

pllim:
I guess it is quite impossible to test `coverage` here. I will undo the test changes and try to test this over at `astropy` instead.

pllim:
pre-commit.ci autofix

pllim:
Okay... autofix didn't do anything and I failed to see how my diff would trigger any such failures.

```
isort....................................................................Failed
- hook id: isort
- files were modified by this hook
```
```
encode scripts in workflows..............................................Failed
- hook id: encode-scripts
- files were modified by this hook
```

.github/workflows/tox.yml
```diff
@@ -223,9 +223,35 @@ jobs:
...
+          path: coverage
+          pattern: coverage_*
+          merge-multiple: true
+
+      - name: Upload report to Codecov
+        if: ${{ hashFiles('coverage/') != '' }}
         uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+          directory: coverage
```
Cadair, line 255: I wonder if we need to specify all the files we downloaded. Unfortunately it seems the download-artifact action only returns the path and not a list of all the artifact files downloaded.

Cadair:
I see the advantage of this, but it does cause you to loose some fidelity in the codecov reports, as you only get one CI job and you loose all the information about what OS / job it was (I think?).

Can you elaborate on the motivation for this a bit more, is it just codecov being flakey?

pllim:
> cause you to loose some fidelity in the codecov reports

Can you please elaborate on this? Each report would be uniquely named when uploaded as artifacts. They will all be downloaded by artifacts, and then re-uploaded to codecov all together. I don't see it being different from what workflow is already doing except the upload to codecov is a separate step.

> motivation

@bsipocz asked for a solution to https://github.com/astropy/astropy/issues/16379

Cadair:
> I don't see it being different from what workflow is already doing except the upload to codecov is a separate step.

I think the difference (and it's not really a deal breaker) is that codecov looses the GHA metadata about which report was from which run?

Anyway, I am happy to merge it, but from the astropy test PR I am unconvinced it actually works?

pllim:
Alternate solution is welcome. We have been using this over at 

https://github.com/spacetelescope/jdaviz/blob/main/.github/workflows/ci_workflows.yml

and

https://github.com/astropy/specutils/blob/main/.github/workflows/ci_workflows.yml

but both workflows only have one coverage job each.

.github/workflows/tox.yml
pllim: Yeah maybe. They have `file` and `files` but the README isn't very clear about usage with wildcards.

pllim:
@Cadair ,  what is this pre-commit silliness?

.github/workflows/tox.yml
pllim: OK, turns out I misunderstood the settings. The actions won't rename the reports, but I managed to separate them out in different sub-directories. The trick was `merge-multiple: false` and luckily codecov uploader is smart enough to find them anyway.

pllim:
I think it works as intended now. Please re-review, @Cadair . Thanks!

Cadair:
What was causing the trouble?

pllim:
Initially, I didn't realize both artifact names actually contains the same `coverage.yml` filenames, and the artifact name does not really reflect the actual filenames at download time, so they were clobbering each other. But I fixed that.

pllim:
@Cadair , are you convinced yet? Anything else I need to do here? Thanks!

pllim:
@Cadair ? @astrofrog ? @ConorMacBride ? 🙏

Cadair (approved):

pllim:
Thanks!

pllim:
Hmm, astropy is picking this up but I don't understand why codecov complains about no token. We have one in repo secrets.

https://github.com/astropy/astropy/actions/runs/9355922312/job/25755726624

Cadair:
I think that's a bug with this. The sunpy builds in particular seem to be attempting to upload the same coverage artifacts more than once: https://github.com/sunpy/sunpy/actions/runs/9352703510/job/25742354670#step:3:6

As you can see in that log it's finding jobs from a different invocation of the workflow (the "test" one) and then downloading them in the "docs" workflow. Do you think you will have time to try and fix it @pllim ?

Cadair:
Ah maybe relatedly, that "docs" job on sunpy's CI has [`pytest: false`](https://github.com/sunpy/sunpy/blob/main/.github/workflows/ci.yml#L56) set, which means it [doesn't upload coverage](https://github.com/sunpy/sunpy/actions/runs/9352703510/job/25741749039), so if the artifact upload / download was working properly it should fail to download any artifacts at all and then [skip the upload to codecov](https://github.com/OpenAstronomy/github-actions-workflows/blob/main/.github/workflows/tox.yml#L254) but that clearly isn't working.

astrofrog:
Yes this is breaking a lot of CI, a lot of my glue builds as well as reproject are failing. I think the artifact name needs to be more unique as the issue seems to happen for me if multiple OSes test e.g. ``py311-test``

pllim:
Not obvious to me how to grab the job ID without doing some Actions YAML output magicks. Is there an easier way?

Issue #189, opened 2024-03-06
Add support for doing code coverage reporting entirely on GitHub


Cadair:
It seems not that complex to upload a coverage report from each job and then combine them in an extra job: https://hynek.me/articles/ditch-codecov-python/ I am finding codecov to be more and more annoying so it might be worth having a second option?

mhvk:
Yes, please!!!

Cadair:
It's unlikely I am getting to this anytime soon if anyone else wants to take a run at it.

pllim:
Looks like pydata-sphinx-theme did the hard work here:

* https://github.com/pydata/pydata-sphinx-theme/pull/1756
* https://github.com/pydata/pydata-sphinx-theme/pull/1809

Though, @weaverba137 also said Coveralls still work (at least for DESI), if that is preferable.

I don't quite remember now why we switched from Coveralls to Codecov in the first place. (Something was broken?)

pllim:
That said, given how much my attempt at #199 broke other repos running CI in different ways, maybe it is better for `astropy` to also ditch the reusable workflow here for CI and only keep it for wheels. That way, we only have to worry about ditching codecov in a way that works for us.

astrofrog:
Why don't we just try and figure it out in the reusable workflows then we would get it to work in all coordinated packages etc too for free?

pllim:
It would be a breaking change. I see very specific codecov settings.

Cadair:
There's a generic `coverage` key: https://github-actions-workflows.openastronomy.org/en/latest/tox.html#coverage so we can add a `coverage: github` which people can opt-in to. I don't think this has to be a breaking change.

Cadair:
From the coverage.py docs I just found this: https://github.com/Bachmann1234/diff_cover we could use that to fail the job based on the patch coverage rather than the whole project.

zacharyburnett:
here's a usage of `diff-cover` (assuming the coverage files are uploaded as artifacts that can be downloaded) 
https://github.com/spacetelescope/stpipe/pull/160/files#r1653419802
```yaml
  report_diff_coverage:
     needs: [ test ]
     if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/setup-python@v5
       - run: pip install diff-cover coverage
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - run: pip install .
       - uses: actions/download-artifact@v4
[...]

PR #206, opened 2024-06-05
Revert splitting out of coverage step

astrofrog:
This reverts https://github.com/OpenAstronomy/github-actions-workflows/pull/199 as it is causing CI to fail in e.g. glue and reproject and other projects. We should make sure we test any follow-up repeat PR more extensively to make sure there are no more issues.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -223,38 +223,9 @@ jobs:
         with:
           paths: "**/results.xml"
 
-      - name: Upload coverage to artifacts
+      - name: Upload to Codecov
         # Even if tox fails, upload coverage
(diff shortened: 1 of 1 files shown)


pllim:
Should have at least kept the `fail_ci_if_error: true` but I don't feel like opening PR here anymore for this issue as I have other things to take care of for now.

astrofrog:
Sorry I just did a plain revert as I was in a rush to fix things, will see if I can add that back later.

pllim:
Might take longer than you expect... I got inexplicable token error over at https://github.com/astropy/astropy/issues/16535

Maybe not worth it and should spend the energy to get rid of codecov altogether... See https://github.com/OpenAstronomy/github-actions-workflows/issues/189

PR #207, opened 2024-06-09
add option to generate test coverage report without codecov

zacharyburnett:
using the guide here: https://hynek.me/articles/ditch-codecov-python/

This does not create coverage diffs or report to PRs, just creates a markdown summary table


Cadair:
I tried this out on sunpy and it seems to explode. https://github.com/sunpy/sunpy/pull/7668

Cadair:
It seems that it's not picking up the coverage.xml to upload here: https://github.com/sunpy/sunpy/actions/runs/9445798095/job/26014155532?pr=7668#step:14:16 :thinking:

.github/workflows/tox.yml
```diff
@@ -229,3 +229,31 @@ jobs:
         uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+
+      - name: Upload coverage data
+        if: ${{ (success() || failure()) && contains(matrix.coverage, 'github') && matrix.pytest == 'true' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-data-${{ matrix.os }}-${{ matrix.toxenv }}
```
Cadair, line 237: @astrofrog is this a sufficiently unique key?

.github/workflows/tox.yml
```diff
@@ -229,3 +229,31 @@ jobs:
...
+        with:
+          pattern: coverage-data-*
+          merge-multiple: true
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - run: python -Im pip install --upgrade coverage[toml]
+      - run: python -Im coverage combine
+      - run: python -Im coverage report --format=markdown >> $GITHUB_STEP_SUMMARY
+
```
Cadair, line 259: I think it would be nice to add the ability to fail this job based on either total coverage % or diff coverage %.

Cadair (commented):
Thanks so much for taking a swing at this!

pllim:
tl;dr -- this does not generate *patch* coverage, right?

Cadair:
Not yet, but I found a way to make it.

zacharyburnett:
> tl;dr -- this does not generate _patch_ coverage, right?

> Not yet, but I found a way to make it.

For patch coverage we'd need to store the coverage from `main` somewhere, to compare it to; some people I've seen online have done that by storing it in a hidden page in the repository's wiki, reading and writing it with a github token with `${{ secrets.GITHUB_TOKEN }}`

pllim:
I am trying this out too at https://github.com/spacetelescope/acstools/pull/205 and it still crashes:

https://github.com/spacetelescope/acstools/actions/runs/9463279566/job/26068438067?pr=205

zacharyburnett:
> I am trying this out too at [spacetelescope/acstools#205](https://github.com/spacetelescope/acstools/pull/205) and it still crashes:
> 
> https://github.com/spacetelescope/acstools/actions/runs/9463279566/job/26068438067?pr=205

oh whoops, I fixed the artifact issue:
image
but didn't pass any filenames to `coverage combine` 😅 
image

pllim:
Getting late here. @zacharyburnett , feel free to restart the CI over at https://github.com/spacetelescope/acstools/pull/205 to pick up new changes from this PR. Thanks!

pllim:
Still not working

```
Run mv .coverage .coverage.ubuntu-latest-py39-test-alldeps
mv: cannot stat '.coverage': No such file or directory
```

pllim:
Ah, looks like the test failed, so no coverage file was written. Maybe have the `mv` command skip if no coverage file is found.

pllim:
Even when CI passes, it failed with the same error.

`Coverage XML written to file coverage.xml`

`mv: cannot stat '.coverage': No such file or directory`

zacharyburnett:
it looks like `--cov-report=xml` is what's creating the `coverage.xml`; however, it **should** still be creating a `.coverage` database file... I'll so some local testing to see what files it spits out
https://github.com/spacetelescope/acstools/actions/runs/9463279566/job/26118526244#step:10:105
```shell
py312-test-devdeps: commands[2] /home/runner/work/acstools/acstools/.tmp/py312-test-devdeps> pytest --pyargs acstools --cov-report=xml --cov=acstools /home/runner/work/acstools/acstools/doc --cov --remote-data -v
```

zacharyburnett:
running `pytest --pyargs acstools --cov-report=xml --cov=acstools --cov --remote-data -v` locally does appear to create the `.coverage` file:
```
❯ ls -a
.							CITATION.md
..							CODE_OF_CONDUCT.md
.bandit.yaml						LICENSE.md
.coverage						MANIFEST.in
.coverage.urelialia2022.local.stsci.edu.35749.XCMKciAx	README.rst
.git							__pycache__
.github							acstools
.gitignore						acstools.egg-info
.mypy_cache						conftest.py
.pytest_cache						coverage.xml
.readthedocs.yaml					doc
.tmp							pyproject.toml
.tox							tox.ini
```

pllim:
What is the difference between `.coverage` and `coverage.xm`?

zacharyburnett:
> What is the difference between `.coverage` and `coverage.xml`?

I'm not intimately familiar with it, but as far as I know `.coverage` is a SQLite database file with support for `parallel-mode` (storing coverage for multiple machine configurations / runs) while `coverage.xml` is report-focused and so is expected to only contain results from a single run

pllim:
Tried to debug at #209 but no luck. I see the coverage ran but the the upload failed: https://github.com/spacetelescope/acstools/actions/runs/9483014271/job/26134042811?pr=205

zacharyburnett:
> Tried to debug at #209 but no luck. I see the coverage ran but the the upload failed: https://github.com/spacetelescope/acstools/actions/runs/9483014271/job/26134042811?pr=205

could you also add `-a` to the `ls`?

pllim:
Ah, yes, added. New log: https://github.com/spacetelescope/acstools/actions/runs/9483014271/job/26138000458?pr=205

zacharyburnett:
testing done at https://github.com/spacetelescope/stpipe/actions/runs/9618208358
[image](https://github.com/spacetelescope/stpipe/actions/runs/9617692573/attempts/7#summary-26531428357)

pllim:
I see error in https://github.com/spacetelescope/stpipe/actions/runs/9618208358/job/26531763051 (despite the green status). Is this expected?

`##[error]Process completed with exit code 1.`

.github/workflows/tox.yml
```diff
@@ -209,8 +209,7 @@ jobs:
         run: python -m pip install --upgrade tox ${{ matrix.toxdeps }}
         shell: sh
 
-      - run: python -m tox -e ${{ matrix.toxenv }} ${{ matrix.toxargs }} -- ${{ matrix.pytest_flag }} ${{ matrix.posargs }}
-        shell: sh
+      - run: tox -e ${{ matrix.toxenv }} ${{ matrix.toxargs }} -- ${{ matrix.pytest_flag }} ${{ matrix.posargs }}
```
pllim, line 212: Why this change?

.github/workflows/tox.yml
```diff
@@ -229,3 +228,42 @@ jobs:
         uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+
+      - if: ${{ (success() || failure()) && contains(matrix.coverage, 'github') && matrix.pytest == 'true' }}
+        run: |
+          mv .coverage .coverage.${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
```
pllim, line 234: So, this step appears to use `runs-on: ${{ matrix.os }}` above. What if someone tries to do this coverage stuff on Windows?

tools/tox_matrix.py
```diff
@@ -26,9 +26,26 @@
 @click.option("--runs-on", default="")
 @click.option("--default-python", default="")
 @click.option("--timeout-minutes", default="360")
-def load_tox_targets(envs, libraries, posargs, toxdeps, toxargs, pytest, pytest_results_summary,
-                     coverage, conda, setenv, display, cache_path, cache_key,
-                     cache_restore_keys, artifact_path, runs_on, default_python, timeout_minutes):
+def load_tox_targets(
```
pllim, line 29: Style formatting is out of scope, no?

.github/workflows/tox.yml
zacharyburnett: oh good catch, that was a debugging step that I thought I had undone

.github/workflows/tox.yml
zacharyburnett: `runs-on` accepts the name of the Docker image (most commonly `ubuntu-latest`, `macos-latest`, or `windows-latest`), but we don't necessarily want that in the coverage combination; `runner.arch` will be `Linux`, `macOS`, or `Windows`

pllim: `mv` command doesn't exist on Windows, does it?

zacharyburnett: it does as an alias in powershell, which I believe is the default for Windows runners:
https://learn.microsoft.com/en-us/powershell/module/Microsoft.PowerShell.Management/move-item?view=powershell-7.4#notes

However to be safe I can set the shell to `bash` here

zacharyburnett:
> https://github.com/spacetelescope/stpipe/actions/runs/9618208358/job/26531763051

yes it should be, I tested with a run with no coverage at all to make sure it didn't adversely affect the step with a false failure

pllim:
I think @astrofrog mentioned elsewhere that would be nice if there can be tests for this, but not sure how to add them.

.github/workflows/tox.yml
```diff
@@ -229,3 +229,43 @@ jobs:
...
+          submodules: ${{ inputs.submodules }}
+          ref: ${{ inputs.checkout_ref }}
+      - uses: actions/download-artifact@v4
+        with:
+          pattern: .coverage.*
+          merge-multiple: true
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+        continue-on-error: true
```
pllim, line 265: I don't know how I feel on so many usage of `continue-on-error: true` in this workflow. I would want to know if a step errored. And I'd rather a step not run when it is not supposed to run using `if` instead of throwing error.

pllim: Otherwise, when a real error happens, it is hard to debug because everything is green no matter what.

zacharyburnett: that is fair; the ideal solution would be to run this job conditionally (`if: `) only when `matrix.coverage` contains `github`; then we could remove these `continue-on-error`s.

However, I'm not sure how to actually do that. I tried doing `if: inputs.coverage == 'github'`, but `inputs.coverage` to this workflow was always blank and did not reflect the contents of the generated matrix.

pllim: Where did you put the `if`? We can put this at the job level, right, since this whole "report coverage" is irrelevant for other modes?

For the combining part, maybe can skip if no coverage files found after download (maybe can use `hashFiles`, see https://stackoverflow.com/questions/71336204/github-action-check-if-a-file-already-exists), and maybe the last 3 Python commands can be just one step.

zacharyburnett: oh, good point! I put the `if` at the job level, but we could put it in the individual steps just as easily

pllim:
BTW I think this fix #189 but only a maintainer can add that to the original post for auto-close to take effect. cc @Cadair

pllim:
@zacharyburnett is this ready for re-review or you still working on it? Thanks!

zacharyburnett:
> @zacharyburnett is this ready for re-review or you still working on it? Thanks!

This is ready now, thanks!

pllim:
Alas, still cannot get this to work with acstools. What am I doing wrong there?

https://github.com/spacetelescope/acstools/pull/205

https://github.com/spacetelescope/acstools/actions/runs/9669917380/job/26677517727?pr=205

zacharyburnett:
> Alas, still cannot get this to work with acstools. What am I doing wrong there?
> 
> [spacetelescope/acstools#205](https://github.com/spacetelescope/acstools/pull/205)
> 
> https://github.com/spacetelescope/acstools/actions/runs/9669917380/job/26677517727?pr=205

I'm not sure, this seems to be correct but the `.coverage` file isn't being created (only `coverage.xml`)

pllim:
A lot of CI only outputs `coverage.xml`. I think this case also has to be handled.

zacharyburnett:
> A lot of CI only outputs `coverage.xml`. I think this case also has to be handled.

It was my belief that `coverage` always creates a `.coverage` file, and `coverage.xml` is an additional report file generated with `coverage xml` or `--cov-report=xml`; however, I guess that's wrong

zacharyburnett:
@pllim would you mind rerunning that job now that 1f4a851 is pushed? It should pick it up

pllim:
Here is the updated error log https://github.com/spacetelescope/acstools/actions/runs/9669917380/job/26717352610?pr=205

pllim:
Maybe the coverage stuff is inside `.tox`

zacharyburnett:
> Maybe the coverage stuff is inside `.tox`

ah! You're using `pyargs`; perhaps that is causing this issue by moving the working dir to somewhere else

pllim:
`astropy` also uses `pyargs`, so it is not uncommon. See https://github.com/astropy/astropy/blob/5519605a3b6ec71e9ddb4c6827b8ac1658647f58/tox.ini#L121-L122

zacharyburnett:
looks like that fixed it https://github.com/spacetelescope/acstools/actions/runs/9683845227/attempts/1#summary-26720539461 thanks for being so patient! 😄

zacharyburnett:
the failing `test_conda` tests are fixed by https://github.com/OpenAstronomy/github-actions-workflows/pull/123

pllim:
Awesome, thanks! Now that the simple case of acstools works, I plan to open an experimental PR to `astropy` with this PR branch, or feel free to do it if you have time now as I cannot until later today.

zacharyburnett:
> Awesome, thanks! Now that the simple case of acstools works, I plan to open an experimental PR to `astropy` with this PR branch, or feel free to do it if you have time now as I cannot until later today.

sure thing, here: https://github.com/astropy/astropy/pull/16619

I don't have permissions to label that PR though

.github/workflows/tox.yml
```diff
@@ -229,3 +229,50 @@ jobs:
...
+        with:
+          python-version: "3.12"
+      - if: steps.check_downloaded_files.outcome == 'success'
+        name: generate coverage report
+        run: |
+          python -Im pip install --upgrade coverage[toml]
+          python -Im coverage combine
+          python -Im coverage report -i -m --format=markdown >> $GITHUB_STEP_SUMMARY
+      - if: steps.check_downloaded_files.outcome == 'success'
+        uses: actions/upload-artifact@v4
```
pllim, line 275: What is this upload for?

zacharyburnett: this is to store `.coverage` for each run as an artifact, and also pass it to subsequent jobs

pllim:
@astrofrog , do you want to take this for a spin on your repos?

.github/workflows/tox.yml
```diff
@@ -238,3 +238,50 @@ jobs:
         uses: codecov/codecov-action@015f24e6818733317a2da2edd6290ab26238649a  # v5.0.7
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+
+      - if: ${{ (success() || failure()) && contains(matrix.coverage, 'github') && matrix.pytest == 'true' }}
+        name: add run info to coverage filename
+        run: mv $(find ~ -name .coverage) .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
```
Cadair, line 244: I seem to be having trouble with this line.

I am trying this out on the dkist repo: https://github.com/DKISTDC/dkist/pull/471 and by default tox puts `.coverage` in the `.tmp/py312/` directory. Even if I move the `.coverage` file up into the root repo directory this still doesn't seem to find it.

Should this in fact be this:

```suggestion
        run: mv $(find . -name .coverage) .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
```

i.e. look in the current directory?

zacharyburnett: ah, good catch, it should be using the working directory

Cadair: What I can't figure out is if there's an way of doing the equivalent of this:
```
            rf"--cov-report=xml:${{GITHUB_WORKSPACE}}{sep}coverage.xml ")
```

Where we can set the .coverage path?

zacharyburnett: `coverage` makes a `.coverage` database by default, if you don't specify a `--cov-report` type. You CAN specify `--cov-report=` to be `html xml json lcov annotate` (see https://pytest-cov.readthedocs.io/en/latest/reporting.html), but as far as I can tell the markdown table generator can only ingest the `.coverage` database

Cadair: Yeah, it's the inability to specify the path of the coverage db that seems to be a problem.

Even with the last commit I don't seem to be finding the db file: https://github.com/DKISTDC/dkist/actions/runs/12161554758/job/33916444661?pr=471#step:14:1

Testing this out locally *seems* to work.

.github/workflows/tox.yml
```diff
@@ -238,3 +238,50 @@ jobs:
         uses: codecov/codecov-action@015f24e6818733317a2da2edd6290ab26238649a  # v5.0.7
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+
+      - if: ${{ (success() || failure()) && contains(matrix.coverage, 'github') && matrix.pytest == 'true' }}
+        name: add run info to coverage filename
+        run: mv $(find . -name .coverage) .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
```
Cadair, line 244:
```suggestion
        run: |
            find . -name .coverage
            mv $(find . -name .coverage) .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
```

.github/workflows/tox.yml
```diff
@@ -238,3 +238,52 @@ jobs:
         uses: codecov/codecov-action@015f24e6818733317a2da2edd6290ab26238649a  # v5.0.7
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+
+      - if: ${{ (success() || failure()) && contains(matrix.coverage, 'github') && matrix.pytest == 'true' }}
+        name: add run info to coverage filename
+        run: |
+          find ${{GITHUB_WORKSPACE}} -name .coverage
+          mv $(find ${{GITHUB_WORKSPACE}} -name .coverage) .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
```
Cadair, line 246:
```suggestion
          find ${{ github.workspace }} -name .coverage
          mv $(find ${{ github.workspace }} -name .coverage) .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
```

.github/workflows/tox.yml
zacharyburnett: I vaguely remember running into this issue where the working directory (and thus the `.coverage` file) was somewhere unexpected. Now that I think about it, that may be why I used `find ~`; I wanted to recursively find the `.coverage` file wherever it was

Cadair: I hope you don't mind, but I pushed a couple of experimental commits to make it search from the root workspace dir down. Let's see what happens.

zacharyburnett: no problem!

Cadair: Ok well the plot thickens: https://github.com/DKISTDC/dkist/actions/runs/12161554758/job/33918322027#step:14:1

The find command finds the file, but then it fails to upload?

.github/workflows/tox.yml
```diff
@@ -238,3 +238,52 @@ jobs:
...
+          find ${{ github.workspace }} -name .coverage
+          mv $(find ${{ github.workspace }} -name .coverage) .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
+        shell: sh
+
+      - name: Upload coverage data to GitHub
+        if: ${{ (success() || failure()) && contains(matrix.coverage, 'github') && matrix.pytest == 'true' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
+          path: .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
```
Cadair, line 254: Ok I am getting desperate now:

```suggestion

      - name: Upload coverage data to GitHub
        if: ${{ (success() || failure()) && contains(matrix.coverage, 'github') && matrix.pytest == 'true' }}
        uses: actions/upload-artifact@v4
        with:
          name: .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
          path: **/.coverage
```

.github/workflows/tox.yml
```diff
@@ -238,3 +238,45 @@ jobs:
         uses: codecov/codecov-action@015f24e6818733317a2da2edd6290ab26238649a  # v5.0.7
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+
+      - name: Upload coverage data to GitHub
+        if: ${{ (success() || failure()) && contains(matrix.coverage, 'github') && matrix.pytest == 'true' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: .coverage.${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.toxenv }}
+          path: **/.coverage
```
Cadair, line 247:
```suggestion
          path: "**/.coverage"
```

.github/workflows/tox.yml
Cadair: I tried to go a different way and let the upload action figure it out: https://github.com/DKISTDC/dkist/actions/runs/12161554758/job/33921111684#step:14:16

It did not go well.

Cadair:
pre-commit.ci autofix

neutrinoceros:
thank you ! now https://github.com/zacharyburnett/github-actions-workflows/pull/1 is much easier to read !
FTR, I did iterate overate it on my own fork, so it should be safe to merge, but feel free to run CI on yours first !

neutrinoceros:
I'm going to lay out the design constraints I'm seeing:
- `tox.yml` can only *append* arguments to a command via `matrix.pytest_flag`, which tox redirects as `posargs`
- in order to automatically form a `coverage` command *dynamically*, I need to be able to *prepend* `coverage run -m` before the pytest invoke
- I would also need to somehow be able to add dynamically generated commands to invoke `coverage report` or `coverage xml` ...

This just seems impossible to do in the most general case. I can see why `pytest-cov` solves these issues but I'm frustrated that it's a de facto requirement. My *prefered* solution would be to renounce doing this stuff dynamically and let users define their own commands (using pytest-cov or coverage directly, at their discretion); but of course that'd be a breaking change so now I'm kinda stuck.

Anyway, this problem I'm reporting isn't specific to this PR, it's pre-existing. All things considered, it seems reasonable to preserve the assumption that pytest-cov is needed for the time being, and I can revisit this adjacent issue later. I'll open a new PR-to-PR to restore functionality.

neutrinoceros:
https://github.com/zacharyburnett/github-actions-workflows/pull/2 👀

neutrinoceros:
I don't understand why it's still failing now. I can spend some more time digging it tommorow. In the mean time, I would suggest adding a debug step to run `ls -R` to see exactly where artifacts get downloaded

neutrinoceros:
(my rant about pytest-cov now has its own issue https://github.com/OpenAstronomy/github-actions-workflows/issues/330)

neutrinoceros:
Phew, took me a while to get there, but I finally have something that works again: https://github.com/zacharyburnett/github-actions-workflows/pull/3

important notes: I couldn't, for the life of me, figure out where `pytest-cov` would write coverage data, nor how it overwrites (as the documentation claims it does) coverage's `--parallel-mode`, so I had to resort to cutting the middle man again and invoke coverage directly. The obvious drawback is that users *need* to do the same thing (`coverage run --parallel-mode -m pytest ...`) from their `tox.ini` file to leverage this. But since it's a *new* feature, there's no backward incompatibility, so it should be acceptable, right ?

neutrinoceros:
@Cadair how should we proceed to get CI green again ?

Cadair:
Thanks everyone!

.github/workflows/tox.yml
```diff
@@ -236,3 +236,47 @@ jobs:
...
+          pattern: coverage-data-${{ github.run_id }}-*
+          merge-multiple: true
+
+      - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
+        with:
+          python-version: "3.12"
+      - name: generate coverage report
+        run: |
+          python -Im pip install 'pip>=25.1'
+          python -Im pip install --group covcheck
```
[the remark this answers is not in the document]
Cadair: https://github.com/OpenAstronomy/github-actions-workflows/pull/356

PR #247, opened 2024-11-11
Add support for customising runs-on with publish

### LINKED ISSUE #227: Support running publish workflow on aarch64 runners
GitHub now has support for paid for aarch64 runners, which means [passing labels to `runs-on:`](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/running-jobs-on-larger-runners#using-labels-to-control-where-jobs-are-run). I don't think there is a way we can do this at the moment, looks like we would need to customise the [`get_os` function](https://github.com/OpenAstronomy/github-actions-workflows/blob/main/tools/load_build_targets.py#L38)

Cadair:
fixes #227


Cadair:
@astrofrog @ConorMacBride can we get this landed and then cut a release?

astrofrog (approved):

PR #278, opened 2025-05-19
DEP: upgrade cibuildwheel to 3.0.0

neutrinoceros:
This is an experimental PR to test `cibuildwheel` 3.0.0b1 here (and downstream if all goes well).


neutrinoceros:
downstream testing isn't going well: https://github.com/astropy/astropy/pull/18159

neutrinoceros:
Looks like `numpy` is also having issues with this upgrade. Specifically, flat (non-`src`) packaging layout seems to be the common denominator of broken packages, and it's possible that `cibuildwheel` will un-break this use case before their final release
Follow https://github.com/numpy/numpy/pull/29007 for the time being

PR #281, opened 2025-06-03
Add ability to select free-threaded Python interpreter

astrofrog:
Work in progress, if -freethreaded is included in the tox env, a free-threaded Python interpreter is used.

I considered using e.g. ``py313t-test`` but I think it would be cleared to have a full word in the tox env name, and it also makes it easier for packages to do different things in tox if free-threaded is selected (for example test building but don't run tests)

### DIFF
--- a/docs/source/tox.rst
+++ b/docs/source/tox.rst
@@ -82,6 +82,9 @@ The name of the GitHub Actions job can be changed with the ``name``
 option as shown above. By default, ``name`` will be the name of the tox
 environment.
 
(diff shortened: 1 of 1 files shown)


astrofrog:
Actually ``py313t`` does have special meaning for tox so we should just use that (https://github.com/tox-dev/tox/pull/3526)

Cadair (approved):

astrofrog:
Hmm now I am having second thoughts about whether the failures were in fact related to my change somehow. Will double check.

PR #286, opened 2025-06-22
ENH: add support for PEP 735 dependency-groups via a new workflow input `test_groups`

neutrinoceros:
This will require https://github.com/OpenAstronomy/build-python-dist/pull/21

PR #291, opened 2025-07-06
DEP: pin internal scripts' dependencies with PEP 723 metadata and `uv run`

neutrinoceros:
I noticed that `pip`, `click`, `pyyaml` and `packaging` were not pinned anywhere, which seemed awkward given the amount of effort that already went into making these workflows as secure and reproducible as possible, so I added script-level PEP 723 metadata and used `uv` (to my knowledge, the only tool supporting this standard at the moment) to leverage it. `uv` should be regarded as an implementation detail here, as other tools might emerge (or exist already ?) that allow this streamlined workflow where scripts are their own source of truth.


neutrinoceros:
Thanks ! trying with `pipx run`

neutrinoceros:
It seems to break the same way, so I think I still need to understand what's broken, but I'd like to keep `pipx` as it's simpler to set up in CI (by virtue of being included in `actions/setup-python`)

neutrinoceros:
Ok I think I got it. The mistake was actually from #286 and already showed on main [example logs](https://github.com/OpenAstronomy/github-actions-workflows/actions/runs/15863656656). The fix is thus independent from my work here and I'll split it out into its own PR.

.github/workflows/tox.yml
```diff
@@ -126,7 +126,7 @@ jobs:
       - run: python -m pip install PyYAML click packaging
```
Cadair, line 126: Do we need these installs if we are pipx running?

.github/workflows/tox.yml
```diff
@@ -126,7 +126,7 @@ jobs:
...
       - run: cat tox_matrix.py
       - id: set-outputs
```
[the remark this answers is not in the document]
neutrinoceros: yes, thanks !

.github/workflows/tox.yml
neutrinoceros: indeed we don't. Thank you for catching this !

Cadair (approved):

Cadair:
Thanks a lot @neutrinoceros I think this is lovely improvement, nice to use a new Python feature.

I'll let either @ConorMacBride or @astrofrog take a second look

astrofrog (approved):
Looks good!

PR #299, opened 2025-07-28
Remove macos-13

### LINKED ISSUE #298: BLD: macos-13 going away
https://github.com/search?q=repo%3AOpenAstronomy%2Fgithub-actions-workflows%20macos-13&type=code

https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down

The macOS 13 hosted runner image is closing down, following our [N-1 OS support policy](https://github.com/actions/runner-images?tab=readme-ov-file#software-and-image-support). This process will begin September 1, 2025, and the image will be fully retired on November 14, 2025. We recommend updating workflows to use macos-14 or macos-15.

Cadair:
Fixes #298

### DIFF
--- a/.github/workflows/test_publish.yml
+++ b/.github/workflows/test_publish.yml
@@ -1,4 +1,3 @@
-
 on:
   workflow_dispatch:
   push:
@@ -33,6 +32,8 @@ jobs:
         - cp31?-macos*
         - cp312-macosx_arm64
         - cp313-macosx_universal2
+        - target: cp311-macosx_x86_64
+          runs-on: macos-latest
         - cp3?-win_amd64
         - target: cp312-manylinux_aarch64
           runs-on: ubuntu-24.04-arm


.github/workflows/test_publish.yml
```diff
@@ -1,3 +1,4 @@
+
```
Cadair, line 1:
```suggestion
```

tools/tox_matrix.py
```diff
@@ -144,14 +142,6 @@ def get_matrix_item(env, global_libraries, global_string_parameters,
     else:
         item["python_version"] = env.get("default_python") or default_python
 
-    # if Python is <3.10 we can't use macos-latest which is arm64
-    try:
-        if Version(item["python_version"]) < Version('3.10') and item["os"] == "macos-latest":
-            item["os"] = "macos-13"
```
Cadair, line 150: I thought about not removing this for a while, but actually I think removing it and having the error be "can't find Python 3.9" is better than "huh what why is this on macos-13".

docs/source/publish.rst
```diff
@@ -60,13 +60,6 @@ To not build any wheels:
    targets: ''
 
 For additional configuration extra arguments can be passed by making a target a dictionary.
-An example of this is specifying the runner for a target, such as building macos x86_64 wheels on native x86_64 runners:
```
Cadair, line 63: Anyone got an idea of an example here which would still work?

docs/source/publish.rst
```diff
@@ -60,13 +60,6 @@ To not build any wheels:
...
-
-.. code:: yaml
-
-    targets:
-      - target: cp311-macosx_x86_64
-          runs-on: macos-13
```
pllim, line 69: Do you want to swap out the example instead of removing it altogether?

tools/tox_matrix.py
Cadair: :shrug: lol

PR #322, opened 2025-11-03
FEAT: re-export cibuildwheel's markdown summary

neutrinoceros:
cibuildwheel 3.x has a really nice feature to summaries its output in markdown.
This re-exposes the summary so it's reusable downstream.
It's a bit non-ideal that users need to do extra work to benefit from it, I'd prefer if it was done automatically but I haven't found a way to do that. For now, opening as a draft, as it's missing extra documentation, but I want to make sure I have the best solution before I write it down.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -96,6 +96,10 @@ on:
         required: false
         default: ''
         type: string
+    outputs:
+      cibw_summary:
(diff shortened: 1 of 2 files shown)


neutrinoceros:
this is working as I expect. See https://github.com/OpenAstronomy/github-actions-workflows/actions/runs/19045799811?pr=322
(scroll past the action graph)

PR #326, opened 2025-11-05
dynamically add supported Python versions to tox matrix, drawing from package metadata (`requires-python`)

### LINKED ISSUE #325: dynamically test Python versions based on Python version pins in metadata
Currently, the Python versions defined in our CI workflows are defined by passing a static list of envs to `.github/workflows/tox.yml`. These need to be manually updated when 1) new Python versions are released as latest, and 2) we drop support for old versions of Python.

Possibly as an alternative to `envs:`, we could add a flag (`supported_pythons`?) that does the following:
1. retrieve the oldest and latest major versions of Python supported by the project, based on pins in `pyproject.toml` `project.requires-python`
    - to start, this only needs to consider the metadata of the top-level project, but perhaps we could also scan dependencies? 
2. build a matrix on-the-fly of each major version of Python between the oldest and latest. For instance, if the oldest supported version of the package is Python 3.11, and the latest is Python 3.14, then the test matrix would be 3.11, 3.12, 3.13, 3.14

zacharyburnett:
closes #325 

adds `fill`, `fill_platforms`, and `fill_factors` parameters to the `tox.yml` workflow call. `fill: true` will append toxenvs to the existing `envs:` list, namely the currently supported Python versions, with factors from `fill_factors` and repeated for each platform in `fill_platforms`. 

Python versions are first drawn from the current public Python releases that are not yet end-of-life (pulled from `endoflife.date` API), and then filtered based on the `requires-python` pins in the package metadata, if it exists. 

For instance, setting 
```
with:
  ...
  fill: true
  fill_factors: xdist
  fill_platforms: linux
```
for `jwst` (which currently specifies `requires-python = ">=3.11,<3.14"`) adds the following toxenvs to `envs:`
```
- linux: py313-xdist
- linux: py312-xdist
- linux: py311-xdist
```

To read the metadata, I imported `peppyproject`, a Python project metadata abstraction library I wrote to convert `setup.cfg` to `pyproject.toml` (but it also works well in this use case, as it will input metadata from `pyproject.toml`, `setup.cfg`, or `setup.py`).


zacharyburnett:
tested in https://github.com/spacetelescope/jwst/pull/10003 and https://github.com/spacetelescope/romancal/pull/2048:

[image](https://github.com/spacetelescope/jwst/actions/runs/19178039119)

[image](https://github.com/spacetelescope/romancal/actions/runs/19178031004)

tools/tox_matrix.py
```diff
@@ -37,7 +37,7 @@ def load_tox_targets(envs, libraries, posargs, toxdeps, toxargs, pytest, pytest_
                      cache_restore_keys, artifact_path, runs_on, default_python, timeout_minutes):
     """Script to load tox targets for GitHub Actions workflow."""
     # Load envs config
-    envs = yaml.load(envs, Loader=yaml.BaseLoader)
+    envs = yaml.load(envs.replace("\\n", "\n"), Loader=yaml.BaseLoader)
```
zacharyburnett: here's where the aforementioned `\\n`s are unpacked

zacharyburnett (commented):
is there a better way to do this one part?

.github/workflows/tox.yml
```diff
@@ -146,9 +175,11 @@ jobs:
           --runs-on "${{ inputs.runs-on }}" --default-python "${{ inputs.default_python }}" \
           --timeout-minutes "${{ inputs.timeout-minutes }}"
         shell: sh
+      - run: echo ${{ steps.set-outputs.outputs.matrix }}
```
Cadair:
```suggestion
```

.github/workflows/tox.yml
```diff
@@ -128,13 +143,27 @@ jobs:
...
+      - if: inputs.fill
+        id: supported-pythons
+        run: uv run supported_pythons.py --package-source . ${{ inputs.fill_platforms != '' && format('--platforms {0}', inputs.fill_platforms) || '' }} ${{ inputs.fill_factors != '' && format('--factors {0}', inputs.fill_factors) || '' }}
```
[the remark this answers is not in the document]
zacharyburnett: It looks like we're also ignoring `zizmor` 's warning for the `set-outputs` call to `tox_matrix.py` as well

PR #331, opened 2025-11-13
FEAT: ensure uv is always available from cibuildwheel

neutrinoceros:
leverage a new feature from cibuildwheel 3.3.0 so that uv is unconditionally available. This allows users of this workflow to leverage `build-frontend=build[uv]` and possibly `build-frontend=uv` in the future

xref: https://github.com/pypa/cibuildwheel/pull/2322


neutrinoceros:
note that uv is special-cased for this option, so that it's only actually downloaded if needed. In constrast,`manylinux` and `musllinux` image already include it.

Issue #342, opened 2025-12-12
conda + py314: libmamba Could not solve for environment specs


pllim:
I really want to a grab a C program from conda-forge for this test. The Python version should not have mattered and yet it is not working. Is this a bug in OpenAstronomy workflow? Am I missing something?

`tox.ini`

```ini
[tox]
env_list =
    py{311,312,313,314}
isolated_build = True

[testenv]
description =
    run tests
extras = tests
conda_deps =
    hstcal
conda_channels =
    conda-forge
commands_pre =
    pip freeze
commands =
    pytest -v {posargs}
```

GitHub job:

```yaml
jobs:
  test:
    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@9f1fedda61294df4c004c05519a3fbf3b8e1f32f  # v2.3.1
    with:
      conda: true
      envs: |
        - linux: py314
```

The error:

```
  error    libmamba Could not solve for environment specs
      The following packages are incompatible
      ├─ conda =* * is installable with the potential options
      │  ├─ conda [4.1.11|4.1.12|...|4.8.2] would require
      │  │  └─ python_abi =* *_cp27mu, which can be installed;
      │  ├─ conda [22.11.1|22.9.0|...|4.14.0] would require
      │  │  └─ python >=3.10,<3.11.0a0 *, which can be installed;
      │  ├─ conda [22.11.1|22.9.0|...|25.9.1] would require
      │  │  └─ python >=3.11,<3.12.0a0 *, which can be installed;
      │  ├─ conda [22.11.1|22.9.0|...|4.14.0] would require
      │  │  └─ python_abi ==3.8 *_pypy38_pp73, which requires
      │  │     └─ python =3.8 *_73_pypy, which can be installed;
      │  ├─ conda [22.11.1|22.9.0|...|4.9.2] would require
      │  │  └─ python >=3.8,<3.9.0a0 *, which can be installed;
      │  ├─ conda [22.11.1|22.9.0|...|4.14.0] would require
      │  │  └─ python_abi ==3.9 *_pypy39_pp73, which requires
      │  │     └─ python =3.9 *_73_pypy, which can be installed;
      │  ├─ conda [22.11.1|22.9.0|...|4.9.2] would require
      │  │  └─ python >=3.9,<3.10.0a0 *, which can be installed;
      │  ├─ conda [23.10.0|23.11.0|...|25.9.1] would require
[...]

PR #352, opened 2026-02-20
Fix conda bootstrap solve failures for py314 tox envs

Flamki:
## Summary
- stop pinning bootstrap conda env to `python=${{ matrix.python_version }}` in `tox.yml`
- keep bootstrap env focused on tooling (`conda`, `tox`) so newer tox targets (e.g. `py314`) do not fail during bootstrap solve
- add `py314` conda scenario to `test_tox.yml`
- document conda bootstrap behavior in `docs/source/tox.rst`

## Why
Issue #342 fails before tox runs because bootstrap solve currently forces `conda + python=3.14`, which is not always solvable.

## Validation
- YAML parsing passes for workflow files
- Python compile checks pass for helper scripts
- pre-commit could not be fully executed in this environment due temporary DNS failure fetching hooks


Flamki:
Pushed a follow-up commit (c22de83) to fix the encode-scripts mismatch by normalizing embedded script payloads in .github/workflows/tox.yml to LF-stable values. This should unblock the failing pre-commit.ci - pr check.

Flamki:
@braingram @astrofrog @Cadair @dpshelio 
This should now fix the bootstrap solve failure described in #342 while keeping the bootstrap environment focused on tooling.

All checks are passing — would appreciate your review when you have time.

PR #386, opened 2026-03-26
Update GitHub Actions workflows and tools

contactmohitsilhare-glitch:
- Fix CRLF line endings to LF in tools/*.py
- Regenerate base64 strings in workflows
- Update astral-sh/setup-uv to v7.4.0
- Update uv to 0.11.1
- Unify actions/checkout versions


Cadair:
This diff is not highlighting what's actually changed so isn't helpful in it's current state. Your description also does not indicate a useful fix.

PR #388, opened 2026-04-06
fix incorrect key in supported_pythons.py

zacharyburnett:
the `pyproject.toml` reading code was incorrectly looking for `table` instead of `project`, and `requires_python` instead of `requires-python`

Sorry, I should have tested #325 more thoroughly after vendorizing the TOML read!

tested here: https://github.com/spacetelescope/romancal/pull/2256, successfully picks up `3.11` through `3.13`


Cadair (approved):

Cadair:
I just tagged v2.6.3 with this fix in.

PR #361, opened 2026-03-03
remove `conda` parameter after deprecation period

zacharyburnett:
Closes #354

Continuation of #360; this should probably go into release `2.7.0` or later, to ensure a sufficient deprecation period

PR #391, opened 2026-05-05
replace `conda:` parameter with `conda_packages:`

zacharyburnett:
alternative to #361

replace the `conda:` parameter (which invoked the now-archived `tox-conda`) with `conda_packages:`. 

If the user populates `conda_packages` then the workflow sets up a Mamba environment that `tox` runs within.

To migrate from `tox-conda`, copy `conda_deps` from `tox.ini` to `conda_packages` in the test workflow, and then also remember to list any non-Python Conda packages in `allowlist_externals`:
```ini
allowlist_externals =
    hstcal
```

**this will require a major version bump**


zacharyburnett:
@pllim this should hopefully be a good way forward instead of using `tox-conda`; please let me know your thoughts. I don't think this should go into 3.0.0 at the moment, probably 4.0.0 eventually

pllim:
Ooo thanks! Is there a way to test with this branch over at https://github.com/spacetelescope/wfc3tools ?

zacharyburnett:
Yes! Just make a PR that changes this line:
https://github.com/spacetelescope/wfc3tools/blob/f0eaea52a6f233c9238d45b2dcfc42a1f326a395/.github/workflows/tests.yml#L25
```diff
-    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@2835f0cacddf3f8de198db9afdb5354a5cebe0ef  # v2.6.3
+    uses: zacharyburnett/github-actions-workflows/.github/workflows/tox.yml@install_conda_packages_directly

```

zacharyburnett:
here: https://github.com/spacetelescope/wfc3tools/pull/119

zacharyburnett:
actually @Cadair if we can squeeze another PR into the 3.0.0 release, this one seems to be working well

zacharyburnett:
wait no never mind, I still need to use `matrix.conda_packages` instead of `inputs.conda_packages`

zacharyburnett:
ok, fixed

zacharyburnett:
tested at https://github.com/spacetelescope/wfc3tools/actions/runs/26904804618

Cadair:
This let's you install binaries for use in tox but not libraries right? You couldn't import a package installed like this?

.github/workflows/tox.yml
```diff
@@ -57,11 +57,16 @@ on:
         required: false
         default: ''
         type: string
-      conda:
-        description: Whether to test with conda (deprecated)
+      conda_packages:
```
Cadair, line 45: Given how this is separate from tox, I wonder if it would be best folded into `libraries` like the system package managers?

zacharyburnett: Oh that's a good idea!

Cadair:
I'm in no mad rush for 3.0 fwiw, so happy to get this in.

.github/workflows/tox.yml
zacharyburnett: actually on second thought, maybe we should keep them separate; the conda environment is distinct from the runner and platform, and you can install both

Cadair: I'm still not entirely sure what the limits of this functionality are so hard to judge. Can you use this to install C libraries and have them picked up inside tox?

zacharyburnett: yes, anything in the conda environment is exposed to tox via `allowlist_externals`

Cadair: `allowlist_externals` is for command line tools only no?

zacharyburnett: oh yes you're right. The only usage of it I've seen is for testing `wfc3tools` with `hstcal`

 I'm not sure if C libraries will be exposed, but I assume they should be since Conda modifies that path

Cadair: I'll have to experiment. I still think that this is still a good match for libraries? There's not any meaningful difference between using conda like this and homebrew?

zacharyburnett: I guess I don't really understand how the `libraries` installer works, so I'm fine with pushing this PR off to `4.0` if you think it can be better

Cadair:
*muffled screaming*

.github/workflows/tox.yml
Cadair: I've completely failed to write a test which checks to see if you can load a c library installed with conda or apt inside the tox env. Feel free to have a play if you want.

I think this functionality should be in libraries, I think it's the sensible place for it to be. Note, that with your current implementation I haven't been able to install a conda package on a per-env basis, you can see this in my broken test commit.

zacharyburnett:
for some reason the `clib` test with `conda_packages:` specified 
```yaml
  - linux: libraries-clibrary
    conda_packages: libffcall
```
isn't actually picking up the conda packages and is just setting up Python with `setup-python` instead:
https://github.com/OpenAstronomy/github-actions-workflows/actions/runs/27213866684/job/80350042102?pr=391

Cadair:
Would you mind if we bumped this to a future release while we get on top of it? I'd like to release 3.0 before I go on parental leave?

zacharyburnett:
sure thing, lets bump it to `4.0`

PR #392, opened 2026-05-05
replace `fill:` parameter with logic to expand `py3*` glob syntax in `envs`

zacharyburnett:
**as this change removes a parameter it will require a major version bump**

better solution for #325 than #326

replaces the `fill:` parameter with just specifying `py3*` or `py31*` in the env name to expand into supported Python versions; I should have done this originally in #326 😅 

this change also moves `supported_pythons.py` into `tox_matrix.py`


zacharyburnett:
the automated tests failed yesterday due to YET ANOTHER github outage 🙄 I'll try to rerun them

zacharyburnett:
standby, I need to clean this up a bit and then it'll be ready

zacharyburnett:
there, done

.github/workflows/tox.yml
```diff
@@ -157,27 +142,24 @@ jobs:
...
       - run: cat tox_matrix.py
+      - if: contains(inputs.envs, '*')
+        id: supported-pythons
+        uses: zacharyburnett/supported-pythons@83fea1cbcd1df5494f6846152992a553def0325b # 2.0.4
```
Cadair: Do you want to move this into OpenAstronomy before we do this?

zacharyburnett:
this moves the "retrieve latest Python versions list from the Internet" functionality from `supported_pythons.py` to the `zacharyburnett/supported-pythons` action. I've published this version of `supported-pythons` to the github actions marketplace, so it should be unaffected if I leave STScI or delete my account or whatever; however, we can also move the entire action repo to `OpenAstronomy/supported-pythons` if that's preferable

docs/source/tox.rst
Cadair, line 1: I approve of this PR based solely on the amount of documentation it removes the need for 😆

tools/tox_matrix.py
```diff
@@ -2,16 +2,21 @@
 # requires-python = "==3.12"
 # dependencies = [
 #     "click==8.2.1",
+#     "packaging==25.0",
 #     "pyyaml==6.0.2",
+#     "requests==2.32.5",
+#     "tomli==2.4.0",
```
Cadair: You aren't actually using these now they are in the standalone action right?

.github/workflows/tox.yml
zacharyburnett: oh yes I was just typing this up. I can't initiate a transfer request because I don't have permissions to create a new repo on OpenAstronomy, so you can either give me those permissions or alternatively fork the repo to the OpenAstronomy org; either way is fine with me

zacharyburnett: forking is probably easier anyway

docs/source/tox.rst
```diff
@@ -86,6 +86,13 @@ environment.
 If the Python version includes a ``t`` suffix, such as ``py313t``, then
 a free-threaded Python interpreter will be used.
 
+Additionally, the Python version can include glob syntax (i.e. `py3*`) to expand into all supported versions of Python (respecting ``project.requires-python`` in ``pyproject.toml`` as conforming with PEP621).
```
Cadair: Is the single wildcard glob actually that useful other than as a way to activate auto-fill from the project metadata? I can't really see the use in `py31*` these days for example as who's using 3.9 now.

I'm not sure I'm proposing changing anything here, just musing.

docs/source/tox.rst
```diff
@@ -86,6 +86,13 @@ environment.
...
+
+For example:
+- ``py*`` for all supported versions of Python (from https://endoflife.date)
```
Cadair: Is this strictly true, or is it from your project minimum up to latest release or end of life if no minimum in the project?

docs/source/tox.rst
```diff
@@ -86,6 +86,13 @@ environment.
...
+- ``py3*`` for all supported minor versions of Python 3
+- ``py31*`` for all supported versions of Python between ``3.10`` and ``3.20``
```
Cadair: These seem not so helpful. If we actually want users to be able to sub-select supported python versions with this I think we'll need cibuildwheel style regex-lite version specs which I don't really have the motivation for?

.github/workflows/tox.yml
Cadair: Can you transfer it to me and then I can transfer it into the org?

docs/source/tox.rst
zacharyburnett: well I figured it's easier to understand for consistency of writing the glob

Plus, in the future there might be some wild feature that's only available on Python 3.20 and above, maybe

docs/source/tox.rst
zacharyburnett: If the project defines a minimum in `requires-python` then it'll use that as the minimum, otherwise it'll use whatever is on endoflife.date (currently 3.10)

docs/source/tox.rst
zacharyburnett: ok, do you think we should change it to just accept `py*` then and get the full list from endoflife.date (and then narrow down with `requires-python`)?

docs/source/tox.rst
Cadair: Could  you just reword this a little to make that clear.

tools/tox_matrix.py
zacharyburnett: good catch
```suggestion
```

docs/source/tox.rst
Cadair: I'm not really sure. Either way is fine by me, as long as we are ok with the scope of this.

docs/source/tox.rst
zacharyburnett: added more explanation in 06020ab

docs/source/tox.rst
zacharyburnett: ok I'll keep it in, since it seems to work through testing

.github/workflows/tox.yml
zacharyburnett: sure, I just sent the request

Cadair: Ok. Transfer complete.

Cadair:
I merged #397 and now it needs a rebase sorry :see_no_evil:

Cadair (approved):

PR #393, opened 2026-05-17
DEP: upgrade cibuildwheel to 4.0.0

neutrinoceros:


neutrinoceros:
I recommend upgrading other GHA dependencies as well and publishing every upgrade at once as a new major release.

Cadair:
(This is a moot point as we are doing a major release for #397 anyway)

Why do you think this should be a major release of the workflows? Is there a breaking change in cibw 4?

neutrinoceros:
Several https://github.com/pypa/cibuildwheel/releases/tag/v4.0.0
Most notably, support for Python 3.8 and 3.13t is gone

Cadair:
I'm not sure if in general breaking changes in downstream actions should warrant major bumps for us, that would mean we do major bumps a lot. I think a clear changelog showing which actions get bumped should be enough.

neutrinoceros:
>  breaking changes in downstream actions 

You mean "upstream" I think. I'm not arguing for *every* major change to translate as a major release here, but I do think that the ones that have user-visible impact should.

> that would mean we do major bumps a lot.

Well, version numbers are cheap (virtually free), so I don't really get that argument.

PR #396, opened 2026-06-03
Do not explicitly specify uv version in calls to setup-uv

Cadair:
I'm not entirely sure why these versions were here, I don't think we want to pin it, latest should be fine?

### DIFF
--- a/.github/workflows/test_tox.yml
+++ b/.github/workflows/test_tox.yml
@@ -26,6 +26,8 @@ jobs:
     uses: ./.github/workflows/tox.yml
     with:
       envs: |
+        - linux: py314-inputs-linux
+          python-version: 3.14.5
         - linux: py312-inputs-linux
         - macos: py311-inputs-macos
         - macos: py39-inputs-macos
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -155,7 +155,6 @@ jobs:
     steps:
       - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # 8.1.0
         with:
-          version: "0.10.6"
           enable-cache: false
           ignore-empty-workdir: "true"
       - if: inputs.fill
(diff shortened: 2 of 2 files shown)


zacharyburnett:
IMO it would be better to have reproducibility, since `uv` is still under active development and is not yet at `1.0.0` (just in case a new `uv` version breaks something)

Cadair:
I kind of feel like if we can't trust latest release of uv then that's a good reason not to use it.

The motivation for this is trying to figure out why uv isn't installing the latest bugfix release of 3.14 [e.g. here](https://github.com/sunpy/sunpy/actions/runs/26889151694/job/79310589681#step:5:37).

I think if we are going to pin this, I'd be nice to both have a reminder to update it, and also a way for users to override it.

zacharyburnett:
that's a good point, I haven't seen a case where a `uv` update breaks existing CI, and `setup-uv` is being used in a lot of other projects without pinning the `uv` version, so it seems good to trust

Cadair:
Are you good with merging this and #397 and putting them both out as a 3.0?

zacharyburnett:
> Are you good with merging this and #397 and putting them both out as a 3.0?

Sure, but could we consider #392 as well if we're doing a major version bump?

zacharyburnett (approved):

neutrinoceros:
> The motivation for this is trying to figure out why uv isn't installing the latest bugfix release of 3.14

To add a bit of context here, this is very much expected for any fixed version of uv: a uv binary has a baked-in manifest of Python versions it knows exist and can install, but it cannot learn about new ones, so uv users who only rely on uv itself to provide Python interpreters generally only have access to Python versions that are older than their uv binary.
Upgrading uv is one way to resolve this. Another is to configure uv use python binaries other than its own (e.g. with `UV_NO_MANAGED_PYTHON=1`).

Upgrades are very generally backward compatible for a vast majority of users; or to put it differently, the few breaking changes that intentionally shipped in minor releases are almost always very niche.

If someone is ever bitten by this, we could simply re-expose `uv-version` to provide a way out. In the meantime, having it unpinned is probably the way to go for most (probably all) users.

PR #397, opened 2026-06-03
Coverage reporting refactor and cleanup

Cadair:
fixes #395 and fixes #383

This PR makes the following changes:

1. **Breaking change** you need to set ``setenv = COVERAGE_FILE={toxinidir}/.coverage`` in `tox.ini` if your tests are not run in the `{toxinidir}` directory, so that reporting finds the `.coverage` file.
1. ``coverage: codecov`` no longer requires the use of ``pytest-cov``, the xml coverage report is generated using coverage.py in a separate step to the tox run.
2. We now support OIDC auth to codecov meaning you don't need ``CODECOV_TOKEN`` if using this.
3. `coverage: github` is now tested against both pytest-cov and coverage.py, and as long as the `.coverage` files are in the root directory will work correctly.
4. As discussed in https://github.com/OpenAstronomy/github-actions-workflows/issues/383#issuecomment-4118635316 codecov will now fail the job on error, including no files updated. **This is particularly critical with the breaking change**.


astrofrog (approved):
This seems ok to me! Did you test it out on a real package? If so feel free to merge.

Cadair:
I'm having trouble with it on sunpy: https://github.com/sunpy/sunpy/pull/8665

Cadair:
Fixed sunpy: https://github.com/sunpy/sunpy/actions/runs/26883832859

tox.ini
```diff
@@ -59,11 +59,17 @@ description = verify pep8
 deps = ruff
 commands = ruff check .
 
-[testenv:py3{10,11,12}{,-conda}]
+[testenv:py3{10-16}{,-conda,-tmpdir}]
```
pllim: `16`? Someone is ambitious.

Cadair: It'll cover us for what another year?

pllim: You don't think they will finally be like "you know what, we should slow down, Python is usable enough as is, let's give Python package maintainers a break"?

zacharyburnett: at least the stable API and ABI3 mean we no longer have to build C extension wheels for every new Python version now 😓

.github/workflows/tox.yml
```diff
@@ -300,20 +300,32 @@ jobs:
         with:
           paths: "**/results.xml"
 
+      - name: Generate xml report for Codecov
+        # Even if tox fails, upload coverage
+        if: ${{ (success() || failure()) && contains(matrix.coverage, 'codecov') && matrix.pytest == 'true' }}
```
zacharyburnett, line 311: simplification
```suggestion
        if: always() && contains(matrix.coverage, 'codecov') && matrix.pytest == 'true'
```

.github/workflows/tox.yml
```diff
@@ -300,20 +300,32 @@ jobs:
...
+        run: |
+          # combine will exit 1 if it's a no-op which is ok for us
```
zacharyburnett: should this have `continue-on-error`?

zacharyburnett (approved):

tox.ini
pllim: Until Python 4 anyway 😉

Cadair:
Do people think it's worth throwing a warning about the `COVERAGE_FILE` change if we get to the codecov upload step and there isn't a `.coverage` file to work with?

pllim:
> Do people think it's worth throwing a warning

I would say, yes. Not everyone follows the PRs here or even read the change log.

.github/workflows/tox.yml
Cadair: Well if this fails codecov upload will fail afterwards?

Cadair: oh you mean split it like the github one?

.github/workflows/tox.yml
Cadair: always will run if the workflow is cancelled which isn't what we want.

Cadair:
Ok, done, and failing test added to demo. I think this is ready, I'll put out a 3.0 after this.

docs/source/tox.rst
```diff
@@ -163,13 +163,60 @@ This option has no effect if ``pytest`` is ``false``.
 coverage
 ^^^^^^^^
 
-A space separated list of coverage providers to upload to, either
-``codecov`` or ``github``. Default is to not upload coverage
-reports.
+The coverage option controls how coverage reports are made and processed after the tox job has completed.
+The default is to not upload coverage reports.
+This option has no effect if ``pytest`` is ``false``.
+This option takes a space separated list of coverage providers to upload to, either ``codecov``, ``codecov-oidc`` or ``github``.
```
pllim: Oxford comma! Also, is one of them recommended over the others?

```suggestion
This option takes a space separated list of coverage providers to upload to, either ``codecov``, ``codecov-oidc``, or ``github``.
```

pllim (commented):
I have a comment and a question.

pllim:
So you just gonna leave the `py314-tmpdir_warning` job red?

Cadair:
> So you just gonna leave the `py314-tmpdir_warning` job red?

yep lol

.github/workflows/tox.yml
zacharyburnett: oh good point, if this fails then the next step would fail too anyway

pllim:
:shipit:

PR #398, opened 2026-06-08
DEP: upgrade Github Action dependencies

neutrinoceros:
Apply all non-major upgrades available now.


neutrinoceros:
I intentionally didn't upgrade `codecov-action` yet, because the latest version is a major (7.0.0) and the version notes doesn't contain any useful context, so I didn't want to just run with it, but seeing the state of CI, maybe older versions are just non-functional now ?

Cadair:
It looks like they just backported 7.0 to 6.0.2 so might as well bump it: https://github.com/codecov/codecov-action/releases

Cadair:
Looks like either 6.0.2 or 7.0 should fix it: https://github.com/codecov/codecov-action/issues/1956

Cadair:
This is going to break all users, so I'm going to do this one manually and release one last release in the 2.x series.

Cadair:
Thanks @neutrinoceros

PR #410, opened 2026-07-24
DEP: allow running python helper scripts with any Python 3.12.x patch version

neutrinoceros:
pinning 3.12 exactly leads to issues with the latest versions of `pipx`, presumably since version 1.15.1 (`ubuntu-latest` currently ships 1.16.0)

close #409 

xref: https://github.com/pypa/pipx/pull/1953


neutrinoceros:
validated at https://github.com/astropy/astropy/pull/20122

neutrinoceros:
unrelated linting issues are addressed at #411

neutrinoceros:
Alternatively, we could also pin the version installed through `actions/setup-python`. Let me know if that's preferred !

astrofrog (approved):

zacharyburnett:
this started failing for us as well yesterday afternoon, good to know there's a fix. Are we concerned about the virtualenv tox tests failing?

zacharyburnett (approved):
if the tests pass this should be good

neutrinoceros:
I think the remaining failures are part stylistic (#411) and part due to a an unpinned version of coverage. I can probably address the latter in another PR too

neutrinoceros:
I couldn't figure out the coverage issue yet. It's definitely unrelated though. I suggest merging and releasing what's already fixed if possible.

zacharyburnett:
@braingram brought up a good point, shouldn't this patch version specifier be `~=3.12.0` or `>=3.12,==3.12.*`? 

the docs at https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release suggest that `~=3.12` would be equivalent to `>=3.12,==3.*`

neutrinoceros:
Thank you ! In that case I think I'd rather pin Python itself

neutrinoceros:
... though I would also need to change these lines, in a similar way, so scratch that

neutrinoceros:
and to avoid any confusion around the meaning of `~=` (which clearly wasn't obvious to me), I switched to an explicit interval.

zacharyburnett:
> and to avoid any confusion around the meaning of `~=` (which clearly wasn't obvious to me), I switched to an explicit interval.

don't worry, I also thought it worked like that until brett brought it up lol

PR #412, opened 2026-07-24
DEP: upgrade minimal requirement on `coverage.py` and leverage auto combining

neutrinoceros:
Tentatively fix the error seen in CI at
https://github.com/OpenAstronomy/github-actions-workflows/actions/runs/30098527565/job/89498555974?pr=410


neutrinoceros:
My initial approach doesn't solve the issue. My guess is that `pytest-cov` isn't playing nicely with latests versions of `coverage` and somehow consumes reports in a destructive fashion.

PR #413, opened 2026-07-24
DEP: set a temporary upper bound to coverage (`<7.14.0`)

neutrinoceros:
This is a temporary workaround (hopefully functional) for the issue I'm trying to address more properly in #412


neutrinoceros:
Still not getting the hang of it.

